PR #2, opened 2017-07-07
Update README.md

king6cong:


sagiegurari:
Thanks for the PR. Issue is that the readme file is actually generated from the site docs so the real fix is in the nav and content md files. So any fix to the readme directly would get deleted in the next release.
I updated the site md files as per your fix and regenerated the readme.
Thanks

PR #41, opened 2017-10-25
Feature/allow kcov code exclusions

azriel91:
Hiya, decided to contribute instead of just raise issues :stuck_out_tongue:. This allows users to pass patterns or region markers to kcov for code to ignore during coverage.

They can add the following to their `Makefile.toml`:

```toml
[env]
CARGO_MAKE_KCOV_EXCLUDE_LINE = "unreachable,kcov-ignore"             # your choice of pattern(s)
CARGO_MAKE_KCOV_EXCLUDE_REGION = "kcov-ignore-start:kcov-ignore-end" # your choice of markers
```

The default is empty, so it doesn't affect anyone's existing coverage stats (for better or worse).

I've added `workspace3` in the examples directory which I've tested using `cargo make --no-workspace coverage`.

I'm not sure how the website code is generated, so this branch hasn't updated the `docs` directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -913,6 +913,14 @@ In order to run coverage in a workspace project and package all member coverage 
 cargo make --no-workspace workspace-coverage
 ````
 
+If you are using `kcov`, you may declare the following environmental variables in `Makefile.toml` to specify lines or regions of code to ignore:
+
+```toml
+[env]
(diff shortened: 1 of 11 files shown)


sagiegurari:
Thanks a lot :+1: 
Once all the CI builds finish, I'll merge it in.
I'll update the docs afterwards (basically the README you updated is auto generated from other files so I'll copy your text to there).

sagiegurari:
Merged. Thanks :+1:

Issue #50, opened 2017-12-26
Make test file pattern configurable and not force test_* naming convention.


sagiegurari:
Based on @azriel91 feedback in #44

sagiegurari:
Users can now override the test_* pattern for integration tests using a new env var.

````toml
CARGO_MAKE_KCOV_INTEGRATION_TEST_FILE_PATTERN = "my_tests_*"
````

**Comment by azriel91** [comment]
hmm, I was thinking of a more "it works out of the box magically", as in given a crate with some integration tests, `cargo make` would be able to discover them without needing them to be changed. The part that I was looking at in #44 was coverage which was looking for integration test executables to feed to kcov, which _might_ be solved for free if we solve the "find test executables, but not the application executable" problem (#51).

**Comment by sagiegurari** [comment]
will try, but rustc doesn't make life easier with all those hashs and stuff.
won't be closing this issue for now, and will try to see.
I do prefer a scripting solution over rust, as kcov is invoked via script so have everything in one place, and I prefer that the rust code you wrote is not inside cargo-make itself as it is not task specific aware.
bottom line, i'll think about this one more.

**Comment by sagiegurari** [comment]
Fixed in 0.9.2, all tests will now run and no configuration is needed or special naming convesions.

**Comment by azriel91** [comment]
I tried out 0.9.2, and the `cargo make --no-workspace workspace-coverage` flow runs every file under `target/debug/deps`, which includes the interactive apps as well as the `<dependency>.d` and `<dependency>.rlib` files. I think we still need the `CARGO_MAKE_CRATE_FS_NAME` check for the virtual workspace. 

A quick scripting solution to reduce the noise could be to run things that match:
```bash
find "${BINARY_DIR}" -type f | grep -v "\.d$\|\.rlib$\|\.so$""
```

Then we still have to solve the "`cargo make --no-workspace workspace-coverage` runs interactive applications" problem, but at least the "run all tests" problem goes away

sagiegurari:
Its actually being executed in target/debug and not target/debug/dep or am I mistaken?
I do add some hash to the end so current solution is ok for you as well. no?

azriel91:
> Its actually being executed in target/debug and not target/debug/dep or am I mistaken?

Yeap that's right, I think I was based off an older `master` when typing that

> I do add some hash to the end so current solution is ok for you as well. no?

It doesn't work for this set up:

```
virtual_workspace
  |- app/app1
  |   |- src/main.rs
[...]

PR #55, opened 2017-12-30
Modified binary directory and test executable discovery for coverage

azriel91:
This should restrict the files run during coverage to the test executables relevant to the workspace.

Issue #50

### DIFF
--- a/src/Makefile.stable.toml
+++ b/src/Makefile.stable.toml
@@ -525,11 +525,6 @@ else
     mkdir -p ./target/coverage
 fi
 
-BINARY_DIRECTORY=target/debug
-if [ -n "$CARGO_MAKE_WORKSPACE_TARGET_DIRECTORY" ]; then
-    BINARY_DIRECTORY="${CARGO_MAKE_WORKSPACE_TARGET_DIRECTORY}/debug"
-fi
-
 KCOV_EXCLUDE_LINE_ARG=""
 if [ -n "$CARGO_MAKE_KCOV_EXCLUDE_LINE" ]; then
     KCOV_EXCLUDE_LINE_ARG="--exclude-line=${CARGO_MAKE_KCOV_EXCLUDE_LINE}"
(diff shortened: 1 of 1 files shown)


sagiegurari:
Thanks, I'll test it as well locally.

sagiegurari:
There are multiple issues with your changes, for example BINARY_DIR isn't defined and the patterns don't work for me that well.
I just pushed a new commit which takes ideas from your changes but also does few more stuff.
I'm not publishing a new version yet, so if you want to test it as well (please do), take the coverage-kcov task definition from the internal toml and put in some external Makefile.toml and run it and tell me if it works good for you.

azriel91:
> BINARY_DIR isn't defined
ah rookie move :man_facepalming:

> take the coverage-kcov task definition from the internal toml and put in some external Makefile.toml and run it and tell me if it works good for you

not sure if I'm just doing something wrong<sup>TM</sup> but for some reason copy pasting the whole `[tasks.coverage-kcov]` and editing the script in my project's `Makefile.toml` doesn't seem to update it, though I know it worked when I overrode `[tasks.workspace-members-coverage.env]`. I do see `[cargo-make] INFO - Using Build File: Makefile.toml` in the output at the beginning of the run. Anyway I just used `cargo install --force --git ...` / `cargo install --force` for local.

---

I just tried out the current master (`17bb1b8`), and it doesn't work with my integration test, but it made me realize a few things:

* `find` wants `-maxdepth 1` to come before `-type f`
* In `cargo make --no-workspace coverage`, the `coverage-kcov` task is run per workspace member, which means using the `find` to choose what executables to run doesn't make sense &mdash; tests from every other crate are run under kcov, whereas we just need to run the executables for the current crate.

    Side note, this causes test binaries that expect the working directory to be the crate directory to fail

Thinking a bit more, instead of using the `find` to find binaries, we run `cargo test` under kcov, which gives these benefits:

* We don't have to figure out what binaries belong to the current crate we are collecting coverage for
* If the binary names change, it doesn't matter to us because cargo handles what binaries to run

That makes the coverage flow work a lot better for my repository, with a caveat:

coverage_unmerged_vs_merged

The last binary that is executed will have the overarching stats on the kcov index page, but I think that problem may already have existed anyway. The `[merged]` stats are still correct, and that's the main one I look at anyway so I'm not fussed.

azriel91:
I pushed a change to this branch, hopefully it works for you :crossed_fingers:

sagiegurari:
That's a really awesome idea. I'll test it tomorrow and if works good, i'll merge it in.
I'll run few tests with standalone, workspace and without dependencies and see if everything checks out.

sagiegurari:
Really tried to make this one work, but with no success.
I either get 0 coverage or in other cases get core dumps: 

```console
> kcov --include-pattern=/projects/rust/src/ target/coverage cargo test

error: process didn't exit successfully: `rustc -vV` (signal: 11, SIGSEGV: invalid memory reference)
````

Which is strange since running the rustc -vV works good, so it doesn't play well via kcov

````console
> rustc -vV

rustc 1.24.0-nightly (3bee2b44c 2017-12-16)
binary: rustc
commit-hash: 3bee2b44cfd5610ad0346f0179602d91a8dedfd0
commit-date: 2017-12-16
host: armv7-unknown-linux-gnueabihf
release: 1.24.0-nightly
LLVM version: 4.0
````

azriel91:
Ah sadness, seems to be https://github.com/SimonKagstrom/kcov/issues/153 / https://github.com/SimonKagstrom/kcov/issues/212. Does it work if you use `kcov --verify <rest of args>`?

azriel91:
Just retried this, running `cargo test` under kcov doesn't actually work, the coverage stats I had were false metrics from a non-clean `target/debug` directory :sob:

azriel91:
Linking related issues:

* https://github.com/rust-lang/cargo/issues/1924
* https://github.com/rust-lang/cargo/issues/3670

There is [`cargo metadata`](http://doc.crates.io/external-tools.html), but I'm not keen on adding yet another dependency that requires compilation to CI

sagiegurari:
I subscribed to both issues in cargo. I think once resolve we should revisit this, but until than I'm not sure there is much we can do.

azriel91:
cool I'll close this episode, and do something to make the build happy in the meantime. Thanks a lot for your efforts!

PR #57, opened 2018-01-02
Add GitLab CI documentation

worr:


sagiegurari:
That's awesome, thanks. Never used gitlab ci before but will definitely help others :+1:

PR #58, opened 2018-01-02
Re-order sed commands for BSD support

worr:
BSD sed does not support placing flags after positional arguments.


sagiegurari:
Thanks :+1:

Issue #59, opened 2018-01-03
Update coverage pattern requirements


sagiegurari:
* support user level configuration to allow user level pattern instead of internal to include/exclude tests
* exclude binaries with same name as crate name

Based on comments in #44

sagiegurari:
release new version with the changes from pr #60 with slight changes and fixes.
@azriel91 please pay attention that I changed the env var name a bit and how it is defined by default (using FS crate name).

azriel91:
Cool thank you! :tada:

PR #60, opened 2018-01-04
Issue 59/update coverage pattern requirements

azriel91:
Addresses issue #59. I've tested this on Ubuntu 17.10 with zsh.

The trickiest part of this was actually getting that double expansion to work; can you check that it works on bash?

I just realized overriding the variable to run my integration test doesn't work for the same reason using `cargo test` didn't work (since `kcov` has to wrap the actual binary call, hah). Anyway the performance improvement is cool :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -937,7 +937,7 @@ In order to run coverage in a workspace project and package all member coverage 
 cargo make --no-workspace workspace-coverage
 ````
 
-If you are using **kcov**, you may declare the following environment variables in your Makefile.toml to customize the coverate task:
+If you are using **kcov**, you may declare the following environment variables in your Makefile.toml to customize the coverage task:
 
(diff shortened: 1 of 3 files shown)


sagiegurari:
Thanks :+1: 
I'll check it out later today.

sagiegurari:
I think instead of:

````toml
CARGO_MAKE_TEST_BINARY_FILTER = "${CARGO_MAKE_CRATE_NAME}-[a-z0-9]*$"
````

You need to:

````toml
CARGO_MAKE_TEST_BINARY_FILTER = "${CARGO_MAKE_CRATE_FS_NAME}-[a-z0-9]*$"
````

I'll merge it in and do few more changes myself on top.
Thanks for keeping up on this item 👍

Issue #56, opened 2018-01-01
 Overriding default tasks non-trivial/annoying.


Cokemonkey11:
#### Problem Description

Creating a `default` task in project's `Makefile.toml` still pulls in dependencies from global default.

More generally, the global defaults are quite heavyweight and overprescriptive.

#### Code Example

````toml
$ cat Makefile.toml
[tasks.default]
dependencies = ["build"]

[tasks.build]
command = "cargo"
args = ["build", "--release"]
dependencies = []
````

#### Error Stack/Info (if any)

```

$ cargo make
[cargo-make] INFO - cargo-make 0.9.2
[cargo-make] INFO - Using Build File: Makefile.toml
[cargo-make] INFO - Task: default
[cargo-make] INFO - Setting Up Env.
[cargo-make] INFO - Running Task: init // ?? Why?
[cargo-make] INFO - Running Task: pre-format
[cargo-make] INFO - Running Task: format-stable
[cargo-make] INFO - Execute Command: "cargo" "fmt" "--" "--write-mode=overwrite" "--force"
Unrecognized option: 'force'.
usage: rustfmt [options] <file>...

Options:
    -h, --help          show this message
    -V, --version       show version information
    -v, --verbose       show progress
        --write-mode [replace|overwrite|display|diff|coverage|checkstyle]
                        mode to write in (not usable when piping from stdin)
        --skip-children
                        don't reformat child modules
        --config-help   show details of rustfmt configuration options
        --config-path [Path for the configuration file]
                        Recursively searches the given path for the
                        rustfmt.toml config file. If not found reverts to the
                        input file path

[cargo-make] ERROR - Error while executing command, exit code: 1
[cargo-make] WARN - Build Failed.
```

sagiegurari:
You just need to understand how to extend tasks and than it's pretty simple once you grasp the idea.

cargo make comes with the following default task definition:

````toml
[tasks.default]
description = "Default task points to the development testing flow"
alias = "dev-test-flow"
````

When you defined the following:

````toml
[tasks.default]
dependencies = ["build"]
````

It doesn't **replace** the **default** task definition, it **extends** it, so cargo-make sees actually:

````toml
[tasks.default]
description = "Default task points to the development testing flow"
alias = "dev-test-flow"
dependencies = ["build"]
````

[...]

PR #63, opened 2018-01-22
Makefile.stable/kcov: remove quotes from args

steveeJ:
With the quotes, if either of KCOV_EXCLUDE_{LINE,REGION}_ARG is empty it
causes one or two empty strings to be passed as arguments to kcov, which
will lead to an error.


sagiegurari:
Thanks, I'll verify this today and push

steveeJ:
You can see an instance of this error in the following log: https://travis-ci.org/wawium/wawium/jobs/331757735#L807-L861

sagiegurari:
I merged and now pushing another fix.
the whole reason you had to do the code change and why you define all those stuff in your makefile is due to a bug I pushed 2-3 days ago as part of #56 
with the fix i'm pushing now (and releasing a new version), you can remove your makefile and everything should work.

PR #79, opened 2018-02-26
Use IndexMap instead of HashMap

### LINKED ISSUE #78: Incorrect composite environment variables resolution order
It seems that when an `[env]` section is declared with some composite variables, the variables are *sometimes* not resolved correctly. It looks like this happens due to the incorrect resolution order. Looking at the code it seems that it iterates over a hashmap (which uses arbitrary ordering) and this may be a potential reason for this issue.

For example with:
```toml
[env]
FOO = "bar"
BAR = "${FOO}/foo"
```

Sometimes the `BAR` variable is rendered as `${FOO}/foo` (as if the `FOO` variable was not resolved).

kostko:
Fixes #78 

This makes the evaluation order of environment variables and possibly other things deterministic (based on insertion order).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ serde = "^1.0.27"
 serde_derive = "^1.0.27"
 shell2batch = "^0.1.12"
 toml = "^0.4.5"
+indexmap = { version = "^0.4.1", features = ["serde-1"] }
 
 [features]
 default = []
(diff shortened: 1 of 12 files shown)


sagiegurari:
Thanks a lot for the implementation.
Can you please do the minor code change I requested so the build will pass?

kostko:
> Basically cache.toml should not be changed

Fixed, thanks I didn't see that change and was wondering why the tests were failing :-)

sagiegurari:
great thanks a lot.
I'll review it more today and run few tests and merge it in.

sagiegurari:
merged. thanks for the fix 👍

PR #127, opened 2018-11-07
Support uploading binaries to GitHub releases

LegNeato:
Not sure this is how you want to do it, or if you want it at the `cargo-make` layer. It might be hard as there are many ways to build a binary for release, but perhaps we do sane defaults and let people override?

I largely cribbed it from my `asciinema-rs` project. You can see the output here: https://github.com/LegNeato/asciinema-rs/releases

The jobs with `DEPLOY=1` are only run when a tag is pushed, which should mean that when you run your current release process the release build artifacts will get kicked off and automatically upload to GitHub.

NOTE: This needs an API token generated to let travis write to github releases. I don't have permission to do so, so that would need to happen before landing.


sagiegurari:
interesting. i need to take a look at this and understand the flow in travis.
want to somehow connect it to my automated release flow i have today in cargo-make

LegNeato:
It is "connected" due to only triggering when a tag is pushed, which looks like it happens in the `cargo-make` publish flow. There are some decent complications putting it in `cargo-make` for the general case (lib vs binaries, what to do about workspaces, etc). I punted on all that :-D

sagiegurari:
@LegNeato you did some amazing work here. I really like it. I know much of the issues has to do with linting of newer lang features which I resolved in branch 0.15.0
I'm thinking maybe we can do 2 things
1. have this PR actually be created on top of 0.15.0 development branch
2. see how we can use cargo-make more to automate things instead of having another shell file like setup_musl.sh

.travis.yml
```diff
@@ -1,26 +1,55 @@
...
-  - 1.27.0
-  - 1.26.0
-  - 1.25.0
-  - 1.24.0
-  - 1.23.0
-  - 1.22.0
-os:
-  - linux
-  - osx
+cache: cargo
```
sagiegurari: why is this needed? for me I prefer to download latest cargo-make each time (its another test in a way)

LegNeato:
Rebased on 0.15.0 branch. Will work on getting the script into `Makefile.toml` next.

.travis.yml
LegNeato: Deleted.

LegNeato:
Ok, two issues:

1. $TARGET doesn't seem to be expanded in the `tasks.build-release` steps.
2. `tasks.binary-distribution-ci-flow.linux` doesn't seem to be getting used on linux. Can you not override dependencies using platform overrides?

Not sure if these are bugs in `cargo-make` or I am doing something wrong.

sagiegurari:
overall looks really good. I really want this one pushed. 

* suggest you use commands and not scripts where ever possible. expansion will happen for command args. also it should be in the form of ${var}, see [docs](https://github.com/sagiegurari/cargo-make#command)
* should work fine actually.
* tasks.build-release - already exists, so not to break backwards, maybe rename yours to tasks.build-release4target or something similar?
* maybe for tasks that are specific for travis we could add a condition: env = { "TRAVIS" = "true" }
* curious, is openssl version 1.0.2l is a must? other versions won't do the trick?
* setup-musl - I would ensure its only for linux. you can alias other platforms to empty task (windows_alias = "empty" for example)

sagiegurari:
I'm working on test changes to support windows travis build to unblock you with that.

sagiegurari:
windows builds will now pass, so you need to merge 0.15 changes to your pr

LegNeato:
One last little bit to fix on Windows. I will fix it tomorrow when I have more time.

.travis.yml
```diff
@@ -1,22 +1,70 @@
 language: rust
-sudo: true
```
sagiegurari: removed by mistake maybe? I think kcov installation would fail without sudo which would mean no code coverage.

sagiegurari:
@LegNeato thanks a lot for all the hard work!!!
I've have wanted to resolve this issue for a long time.
I'll setup the encrypted key soon.

.travis.yml
LegNeato: No, I removed it to test as builds are generally faster on the non-sudo infra. I'll poke around to see if it is needed or not.

sagiegurari:
this looks awesome! I see ```sudo make install`` for kcov works without sudo... makes sense :)
anyhow this is really cool so tell me if anything is missing on your end and if not, i'll merge it in, add the github key and release 0.15.0 with some additional docs

sagiegurari:
you can use secure token:

```
U5ziHSOtcXHe7XX1KdEIIVVbOd/AKMvXquYoS2GJnnil3gxGRarOGkZ/F3iFGEzccXwNPuvhnBwSciVMx/Rsx4gl6ET8bila8T4/AppRhORBM4RBfAweJVZUibDoGzQVjc9Y4Aeo2E2oy8yMNFAh8WLu9tY4v5x/hlgUmffW88j3Mg6EdJb5rMQZ0aqyuTzJm6XvDixQQLaJ1bjONFAG7eRQzfs+TC/wiDXipm2bs/wcLcCl/GeB3KDyT7GH1dkeyChwdVt2rqiQM2N7gwSVKZsiV9ZFcv4SFRdwl/W2FKSAlNeIF68qXA4OAaWQxWIlfHqUk6iYgzCA8XbFIFDgRFjFvLEP5fK8Ul5gsOCZF6D9hSDYXvFvbPEqqaOKJUlxZrOaID8+j64E5gI5cPMGlYG4OPkqRYvx/NsES4OUhkrfr8j2wFAUqBPrBqCOnbp1PUkLFoQGRxy28gudhoHPjfPYcFokNErriwz4HG05UBIeGQzUKLNynxVFrwMN8u8UxOh6TDMmOa+okWLcn0jnZPTTi7zCtz34G4ltE/ikL5ERK0tV/lyvqbI6PXQfpW1mXVWf00pfkW88vXaE/uYlOdHU3FnpcGV4nynpniIvk0VJ78Fp/g2cwxw4QTMb6rIjwxmH0Dl9s1aiU/fNHnm/6vE46dwdOdykj1Uim0XgVmw=
```

sagiegurari:
cool. i'll merge it, update the docs and release 0.15.0 so we can test it live.

PR #137, opened 2018-11-13
Fix windows zip file for deploy

LegNeato:


LegNeato:
This isn't ready / doesn't fix the issue. Guess I'll have to figure out how to get my local windows VM back up, sigh.

PR #141, opened 2018-11-16
Add check-format task to default Makefile

### LINKED ISSUE #138: setup rustup via install_crate for fmt check

D4nte:
`check-format` is added as a default task.

This task can be useful for pre-commit hooks or CI checks.

Let me know if you'd prefer to have it added to the `ci-flow` too.

Resolves #138.

### DIFF
--- a/src/Makefile.stable.toml
+++ b/src/Makefile.stable.toml
@@ -216,6 +216,30 @@ dependencies = [
     "post-format"
 ]
 
+[tasks.check-format-stable]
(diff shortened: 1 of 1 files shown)


sagiegurari:
that's great. can you do a PR on yop of 0.15.2 please where current development is being done.
I already added the check task but in your PR you could add it to the ci flow which is a great ifea.
however in order not to break backwards compatibility, you csn gate it with an env var condition, same as i did for running examples in ci flow

D4nte:
Replaced with #149.

PR #148, opened 2018-11-26
package additional executable without cargo dependency

### LINKED ISSUE #69: package additional executable without cargo dependency
This will allow to use cargo-make without cargo to run tasks that do not have to do with project build.

SamuelMarks:
Closes #69


Cargo.toml
```diff
@@ -40,6 +40,11 @@ serde_derive = "^1.0.80"
 shell2batch = "^0.2.4"
 toml = "^0.4.8"
 
+[[bin]]
+bench = false
+path = "src/main.rs"
+name = "cargo-bin"
```
sagiegurari: i would call it makers (make rust) since I think that point here is to make it a non cargo plugin so better remove cargo from its name. what do you think?

src/cli.rs
```diff
@@ -17,18 +17,17 @@ use crate::runner;
 use crate::types::{CliArgs, GlobalConfig};
 use crate::version;
 
-static NAME: &str = "make";
-static VERSION: &str = env!("CARGO_PKG_VERSION");
-static AUTHOR: &str = env!("CARGO_PKG_AUTHORS");
```
sagiegurari: why this block change is needed? i hate going and manually changing things like version. won't it work with a secondary binary?

src/cli.rs
```diff
@@ -172,85 +165,98 @@ fn create_cli<'a, 'b>(global_config: &'a GlobalConfig) -> App<'a, 'b> {
...
-                    .long("--list-all-steps")
-                    .help("Lists all known steps"),
-            ).arg(Arg::with_name("TASK").help("The task name to execute"))
-            .arg(
-                Arg::with_name("TASK_ARGS")
-                    .multiple(true)
-                    .help("Task arguments which can be accessed in the task itself."),
-            ),
-    )
+    App::new("cargo-make")
```
sagiegurari: haven't tried it, so this would would work for a standalone binary and for a cargo subcommand at the same time?

Cargo.toml
SamuelMarks: Yeah, was chatting about that in the office with Rishab today. How about `runrs`? 📦

SamuelMarks:
Thanks for your review.

Will look it over in detail in our timezone, tomorrow with @rishflab

Cargo.toml
sagiegurari: ya something like that.
There is already a rust project with that name: https://github.com/Xion/runrs but its not published to crates.io so I think it should be ok...

src/cli.rs
rishflab: Not 100% sure, I think this change will no longer work with cargo subcommand. In my opinion it is best to choose whether to make this tool a standalone binary or a cargo subcommand rather than support both.

sagiegurari: don't think for backward compatability we should drop support for old executable and not sure how hard it is to support both.
i'll try it out as well to see complexity

rishflab: ill wait for your decision on whether you want to go with standalone binary, cargo subcommand or both before I spend more time on this

sagiegurari:
I was able to support 2 binaries in parallel.
I'll publish it in a new branch and create a PR so you can review as well.
however, some tests are failing now (those that expect panic using the logger error), I'll have to investigate it.

sagiegurari:
@SamuelMarks @rishflab maybe you can look at #150
I think I'll merge that as it keeps both binaries and allows me to be backward compatible.
do you see any issues there?

PR #149, opened 2018-11-28
Add check-format task to default Makefile

### LINKED ISSUE #138: setup rustup via install_crate for fmt check

D4nte:
`check-format` is added as a default task.

This task can be useful for pre-commit hooks or CI checks.

Let me know if you'd prefer to have it added to the `ci-flow` too.

Resolves #138.

### DIFF
--- a/src/Makefile.stable.toml
+++ b/src/Makefile.stable.toml
@@ -216,6 +216,30 @@ dependencies = [
     "post-format"
 ]
 
+[tasks.check-format-stable]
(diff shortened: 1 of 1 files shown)


sagiegurari:
thanks a lot, but I already added that task.
But if you can, please add the task to the ci flow if and only if a new env var is set to true (similar to bench-conditioned-flow)

sagiegurari:
@D4nte can you checkout my earlier comments please?

D4nte:
Replaced with #161

PR #150, opened 2018-11-28
package additional executable without cargo dependency #69

sagiegurari:


SamuelMarks:
Looks good! - Will test when I get back to the office.

On the naming side, rather than `rsmake` call it `makers` 🤖🤓

PR #156, opened 2018-12-03
Shebang support

### LINKED ISSUE #130: [feat] Replace `script_runner` with explicitly required shebang lines?
#### Problem Description
With the exception of `@shell`, why not just use the [shebang line](https://en.wikipedia.org/wiki/Shebang_(Unix))?

Yes, Rust can be covered also, see [this stackoverflow question](https://stackoverflow.com/q/41322300).

Seems like a much more general solution than your current one.

What do you think 😄 ?

SamuelMarks:
Closes #130


SamuelMarks:
To be explored later, is proper Windows shebang support. Using the `PATH`, listing then filtering by stat of file & executable, we can build our own `/usr/bin/env` for Windows. Whence shebang excludes `env‘, on Windows we'll act as if it was there.

Additionally, we can explore using file extension & [magic](https://docs.rs/magic/0.12.2/magic/#usage-example) to infer shebang.

src/scriptengine/mod_test.rs
```diff
@@ -171,3 +183,18 @@ fn invoke_generic_runner_error() {
...
     assert!(output);
 }
+
+#[test]
+fn invoke_shebang_when_runner_not_specified() {
+    let mut task = Task::new();
+    task.script_runner = None;
+    task.script_extension = Some("py".to_string());
+    task.script = Some(vec![
+        "#!/usr/bin/env python".to_string(),
```
sagiegurari: if python is not available it will fail.
lets change to linux only test and add here user sh and not py

src/scriptengine/mod.rs
```diff
@@ -49,7 +48,17 @@ fn get_engine_type(task: &Task) -> EngineType {
                 }
             }
         },
-        None => EngineType::Unsupported,
+        None => match task.script {
+            None => EngineType::Unsupported,
+            _ => {
+                if task.script_extension.is_some() {
```
sagiegurari: I'm a bit worried about this case.
previously, if you didn't have a runner but you did have a script, I would attempt to execute it using sh/cmd.
now, we go to the generic script runner.
the generic script runner will panic if no shebang is defined, but if its not defined, why can't we fallback like before to sh/cmd?
So i would like a is_supported function to be exposed in the generic runner which would check if a shebang is defined, and that would be checked in this line.
if not, instead of panicking it would basically not run the script and allow the runner.rs to fallback to run it using sh/cmd.

src/scriptengine/generic_script_test.rs
```diff
@@ -5,17 +5,51 @@ use crate::test;
...
+        &vec!["echo hello".to_string()],
+        Some(test::get_os_runner()),
+        test::get_os_extension(),
+    );
+}
+
+#[test]
+fn execute_python_hello() {
+    let script_test = vec![
+        "#!/usr/bin/env python".to_string(),
```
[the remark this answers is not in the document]
rishflab: will do, I think this might also be the reason why I am unable to get these tests to pass

SamuelMarks:
Reopened in https://github.com/sagiegurari/cargo-make/pull/165. Using feature branches from now on.

PR #161, opened 2018-12-09
Add format check and clippy to ci-flow

D4nte:
* env CARGO_MAKE_CLIPPY added
* env CARGO_MAKE_CHECK_FORMAT added

### DIFF
--- a/src/Makefile.stable.toml
+++ b/src/Makefile.stable.toml
@@ -26,6 +26,8 @@ category = "CI"
 dependencies = [
     "pre-ci-flow",
     "pre-build",
+	"conditioned-check-format",
+	"conditioned-clippy",
     "build-verbose",
     "post-build",
     "pre-test",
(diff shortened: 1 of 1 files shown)


sagiegurari:
thanks.
can you prefix both env vars as 'CARGO_MAKE_RUN_' for example CARGO_MAKE_RUN_CLIPPY?
also do not use the pre hooks, I leave those emtpy so users can override them.
instead just add to the ci-flow dependencies, like bench-ci-flow and examples-ci-flow

D4nte:
Comments implemented.

src/Makefile.stable.toml
```diff
@@ -219,6 +228,10 @@ category = "Documentation"
 
 [tasks.pre-build]
 category = "Build"
+dependencies = [
```
sagiegurari: lets remove this block. no longer needed

D4nte: Done locally but forgot to stage it 😖

sagiegurari:
merged. thanks @D4nte

PR #163, opened 2018-12-11
File runner (#3)

### LINKED ISSUE #131: [feat] URL/URI in task, maybe as new `script_runner`?
#### Problem Description
Rather than only supporting inline code, why not have a path parameter? - How about going further, and making that path one of URL or URI (see [RFC for file URI](https://tools.ietf.org/html/rfc8089))?

If URL, allow for a SHA256 sum to specified.

#### Potential
cargo-make becomes a package manager 😮 

---

What do you think? 😃

SamuelMarks:
Closes #131


sagiegurari:
are we sure this is needed as it is right now?
i mean, if i want to run some shell file i could do

```toml
[tasks.runfile]
command = 'sh' # any runner i want
args = [ 'examples/myfile.sh' ]
```
and you can fo a platform override task to put something different for windows.

so I'm not sure what this PR gives more that can't be achieved right now

SamuelMarks:
With the shebang PR, we can write the makefile without knowing the command to execute it with.

sagiegurari:
I'm for the shebang line, but why is that simpler than writing the command?
its less understandable (i didn't understand without reading the docs why both scriptpath and script text are defined) and complex to define.
I'm not sure about this one.

when i was thinking about #131 I thought about something like this

```toml
[tasks.remote_script]
script = { url = "some url" }
```

this allows further extension by adding more future attributes and we instead of writing the script in the makefile, we pull it from somewhere.

however, I don't want wget/curl to be implemented inside cargo-mske, so i only planned to pursue this if such rust library existed.

rishflab:
> its less understandable (i didn't understand without reading the docs why both scriptpath and script text are defined) and complex to define.

I will fix this. Script text should not be defined if script path is defined. I wanted to show what the contents of the remote script would be in both scenarios (shebang define and shebang not defined). I will change it to make it clear.

sagiegurari:
can we redesign the structure to be like i suggested?
and we should only support this if there is a wget/curl like library to actually pull the content of the script.
what do you think?

rishflab:
Yes. I will change it so that it works like this:
> ```toml
> [tasks.remote_script]
> script = { url = "some url" }
> ```

sagiegurari:
great. can you tell me what lib we will be using to pull the contents (after you do some investigating)?
or we could use a native executable if exists (not sure what is available for windows)

rishflab:
Yeh sure. I am looking at [reqwest](https://github.com/seanmonstar/reqwest) and [hyper](https://github.com/hyperium/hyper).

As a result cargo make will support both local files and remote files.
'''toml
[tasks.http_script]
script = { url = "http://bla.bla.org/installstuff.sh" }
[tasks.local_file]
script = { url = "file://Users/bla/folder/installstuff.sh" }
'''

Also, slightly off topic: where/how is cargo test called from cargo make. Cant seem to find cargo test anywhere in this projects Makefile.

sagiegurari:
the external file extends the internal makefile.stable.toml
so everyone, even without an external makefile get a wealth of tasks to use out of the box

rishflab:
I have run into an issue with the proposed syntax to specify a url script. There is no rust struct which fits the desired toml. 

`
#[derive(Debug,Deserialize)]
enum Script {
    String,
    Url(String),
}`

^That won't compile.

If we do something like this
`
[tasks.remote_script]
script.url = "http://scripts.com/hello.sh" }
[tasks.file_script]
script.inline = 
```
echo 'hello'
```
 }
`

We can parse it into the following struct using serde+toml:

`
#[derive(Debug,Deserialize)]
enum Script {
    File(String),
    Url(String),
}`

SamuelMarks:
You could try doing something like: https://serde.rs/string-or-struct.html

rishflab:
> You could try doing something like: https://serde.rs/string-or-struct.html

yep, good find, that should work

sagiegurari:
this has been implemented several months ago a bit differently (script attribute can now hold an object with file attribute for file path), so i think we can close this one.

PR #249, opened 2019-06-06
Update README - installing `cargo-make` when caching `cargo`

### LINKED ISSUE #247: Travis-ci `cache: cargo` and installing cargo-make with `--debug`
Hello,
Again more of a question. Maybe it's good idea to add example with `cache: cargo`, because using the `--debug` will fail when there is this configuration in the `.travis.yml`

elpiel:
Resolves #247

### DIFF
--- a/docs/_includes/content.md
+++ b/docs/_includes/content.md
@@ -1929,6 +1929,17 @@ script:
   - cargo make ci-flow
 ```
 
+When caching `cargo`:
+
+```yaml
+cache: cargo
+script:
+  - which cargo-make || cargo install cargo-make
+  - cargo make ci-flow
+```
+
+*NOTE:* If you want to update cargo-make, then you need to clear the cache manually
+
 If you want to run code coverage and upload it to codecov, also define the following environment variable:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


sagiegurari:
thanks a lot!!!
few comments:

* can you move the debug flag comment to the travis section? i'm not sure if caching is available for every CI provider so it might be a bit confusing? not sure. what do you think?
* can you do the changes in content.md instead of readme.md? the readme is generated by running:
```cargo make generate-readme``` which takes the nav and content md files and also updates some other stuff...
so if you change it directly, it will get lost.

**Comment by elpiel** [comment]
Thanks for the fast response. I will update it. Sadly I don't know if all CIs have caching, I've only worked with Travis & Circle. They both have and I though it's a good idea to mention it in the `Continuous Integration` section instead of each separate CI. I could also add the CircleCi caching example, but for the rest, I cannot promise.
```
jobs:
  build:
....
    steps:
      - checkout
      - restore_cache:
          key: project-cache
.....
      - save_cache:
          key: project-cache
          paths:
            - "~/.cargo"
            - "./target"
```

**Comment by sagiegurari** [comment]
I rebased your PR on top of 0.19.5 branch which is the current development branch...

**Comment by sagiegurari** [comment]
thinking about it a bit, lets just drop the whole first change with the debug flag.
I think having concrete examples for caching for specific providers that users can copy paste from, is good enough.
i'll be happy if you include both travis and circle

**Comment by elpiel** [comment]
I've pushed the changes. For `Circle-ci` I haven't tested the example, as currently I am not using Circle-ci, I might take a look and test it, but it will be definitely next week...

**Comment by sagiegurari** [comment]
it looks great. once you tell me it's tested, I'll merge it in.

**Comment by elpiel** [comment]
I've tested it, the only this was that there has to be space like so:
```
- run:
    command:
```
not 
- run:
  command:
```

sagiegurari:
thanks!!!!

PR #253, opened 2019-06-18
Allow passing an array of argument to test_arg

### LINKED ISSUE #252: Multiple test_args when checking for installed crate
**Features Description**
To test if `cargo-travis` is installed, the user is supposed to run `cargo doc-upload --help`. This doesn't work with `cargo-make` however, because it is impossible to pass multiple arguments (`test_arg` is passed as a single argument).

Simply using `cargo-doc-upload --help` wouldn't work because it exits with an error, leading to `cargo-make` thinking it should update it.

Add a new key to install_crate, `test_args`, an array of arguments to pass to the binary. Example:
`install_crate = { crate_name = "cargo-travis", binary = "cargo", test_args = ["doc-upload", "--help" }`

**Code Sample**
```toml
[tasks.deploy-doc]
workspace = false
install_crate = { crate_name = "cargo-travis", binary = "cargo", test_arg = "doc-upload --help" }
install_crate_args = ["--git", "https://github.com/roblabla/cargo-travis", "--rev", "doc-upload-target"]
description = "Upload this project's documentation on github pages. Should only run on CI."
command = "cargo"
args = ["doc-upload", "--clobber-index"]
```

This will always cause cargo-make to reinstall cargo-travis.

roblabla:
Fixes #252 

Still need to add some tests and docs.

This changes `InstallCrateInfo`'s `test_arg` to take either a string (as is currently allowed) or an array - by using a custom deserializer function. In other words, both of those rules are now allowed:

```toml
[tasks.install-mkisofs-rs]
install_crate = { crate_name = "mkisofs-rs", binary = "mkisofs-rs", test_arg = "--help" }

[tasks.install-cargo-travis]
install_crate = { crate_name = "cargo-travis", binary = "cargo", test_arg = ["doc-upload", "--help"] }
```

### DIFF
--- a/src/lib/installer/crate_installer.rs
+++ b/src/lib/installer/crate_installer.rs
@@ -18,7 +18,8 @@ fn invoke_rustup_install(toolchain: &Option<String>, info: &InstallCrateInfo) ->
             let rustup_component_info = InstallRustupComponentInfo {
                 rustup_component_name: component.to_string(),
                 binary: Some(info.binary.clone()),
-                test_arg: Some(info.test_arg.clone()),
+                // InstallRustupComponentInfo only supports one argument right now.
+                test_arg: info.test_arg.get(0).cloned(),
             };
             rustup_component_installer::invoke_rustup_install(&toolchain, &rustup_component_info)
         }
--- a/src/lib/installer/crate_installer.rs
+++ b/src/lib/installer/crate_installer.rs
@@ -18,8 +18,7 @@ fn invoke_rustup_install(toolchain: &Option<String>, info: &InstallCrateInfo) ->
             let rustup_component_info = InstallRustupComponentInfo {
                 rustup_component_name: component.to_string(),
                 binary: Some(info.binary.clone()),
-                // InstallRustupComponentInfo only supports one argument right now.
-                test_arg: info.test_arg.get(0).cloned(),
+                test_arg: Some(info.test_arg.clone()),
             };
             rustup_component_installer::invoke_rustup_install(&toolchain, &rustup_component_info)
         }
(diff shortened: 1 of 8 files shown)


src/lib/installer/crate_installer.rs
```diff
@@ -18,7 +18,8 @@ fn invoke_rustup_install(toolchain: &Option<String>, info: &InstallCrateInfo) ->
             let rustup_component_info = InstallRustupComponentInfo {
                 rustup_component_name: component.to_string(),
                 binary: Some(info.binary.clone()),
-                test_arg: Some(info.test_arg.clone()),
+                // InstallRustupComponentInfo only supports one argument right now.
```
sagiegurari: why not implement it as well?

roblabla: Right, so the reason I put it off was that InstallRustupComponentInfo has an additional requirement that it must also work when `test_arg` is None, in which case no installation tests are done and cargo-make will just always attempt to install.

I'll see if I can find a clean way to incorporate this logic.

roblabla:
I think it should all work. I had to change test_arg to be a new type, `TestArg`, which implements deserializer. That was the only way I could come up with to make `Option<TestArg>` to be None when test_arg is absent. Specifying a custom deserializer through deserialize_with wouldn't work no matter what I tried :shrug:

sagiegurari:
looks great. once CI is done, i'll merge it in. thanks!!!
i'll probably wait for few changes in shell2batch (your PR and i'll also add that user provided translation as well) and release all together.

sagiegurari:
@roblabla merged. thanks a lot!!!
currently pushed to a dev branch of the next release.

PR #292, opened 2019-09-28
Use IndexMap::swap_remove instead of remove

roblabla:
The remove function is deprecated. Swap_remove should be used instead.

Cargo-make currently fails to compile without this change.

-----------------------

As a side-note, I think cargo-make should commit its Cargo.lock in the repo. The lack of a Cargo.lock means that cloning the repo and building might lead to compilation errors due to newer versions of the dependencies breaking the build (as is the case with index-map here).

### DIFF
--- a/src/lib/descriptor.rs
+++ b/src/lib/descriptor.rs
@@ -38,7 +38,7 @@ fn merge_env(
         let value_clone = value.clone();
 
         if merged.contains_key(&key_str) {
-            let base_value = merged.remove(&key_str).unwrap();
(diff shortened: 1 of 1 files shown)


sagiegurari:
Thanks a lot!

as for the lock, keep in mind that it fails due to harsh linting rules i enabled.
those rules only apply when you indeed clone and compile but not when you install it.
so users are not impacted, only cargo-make developers (like you).
i actually want it to break to make sure i'm using the latest stuff.
in this specific case, index map did a backward comparability break without increasing the versions correctly but it happens :)

PR #298, opened 2019-10-02
Add minimal example for drone.io

### LINKED ISSUE #257: Add example usage with drone.io
Adding documentation example of a working CI build using drone.io (.drone.yml)

dtamai:
This should fix #257 

I tried that configuration file in the `examples/workspace` project and it works.

I didn't add anything about cache because I'm not sure how well it would work when caching the `~/.cargo` folder, and right now I can try only the local runner, so no cache available.


sagiegurari:
thanks a lot. will check it out later today and merge it.

PR #299, opened 2019-10-05
Color configuration by environment variable

emakryo:
This PR introduces a new environment variable `CARGO_MAKE_DISABLE_COLOR, which controls whether output is colorized. It works well with nested `cargo make`, e.g. workspace. Also, a new global configuration `disable_color` is added, which works like setting the env variable. These changes fix the following inconsistent behavior in a workspace (e.g. `example/workspace2`).

Before:
cargo_make_before

after
cargo_make_after

### DIFF
--- a/docs/_includes/content.md
+++ b/docs/_includes/content.md
@@ -2768,6 +2768,9 @@ The following example config.toml shows all possible options with their default 
 # The default log level if not defined by the --loglevel cli argument
 log_level = "info"
 
+# The default configuration whether output coloring is disabled
+disable_color = false
+
 # The default task name if no task was provided as part of the cargo-make invocation
 default_task_name = "default"
 
--- a/examples/cargo-make/config.toml
+++ b/examples/cargo-make/config.toml
@@ -1,4 +1,5 @@
 log_level = "error"
+disable_color = true
 default_task_name = "build"
 update_check_minimum_interval = "daily"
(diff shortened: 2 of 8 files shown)


sagiegurari:
i wonder if NO_COLOR doesn't already disable colors?
as I'm using the colored crate which supports it.
so I'm not sure you need to have cargo-make code changes all.
can you test and tell me?

emakryo:
Thank you for telling NO_COLOR variable. I can use it for that purpose.
But, I think it would be convenient to disable colors using only --no-color flag even with workspaces.
Current behavior seems misleading because all outputs should be colorized with the flag.
How about overriding NO_COLOR variable when the flag is passed?

sagiegurari:
yap. you can set it based on the cli. that sounds like a good idea

sagiegurari:
@emakryo can we set the NO_COLORS based on the cli input in the [setup_env](https://github.com/sagiegurari/cargo-make/blob/master/src/lib/environment/mod.rs#L295) function?
this way, as you said, it will drill down to sub processes in case of workspaces.

sagiegurari:
forget my previous comment, i'm looking through your code and I actually think its great.
we could enhance it to do a bit more but not a must in this specific PR.

sagiegurari:
I checked it and it looks great. i'm merging it to a new dev branch.

PR #305, opened 2019-10-16
Guard against self-referential aliases

phi-gamma:
When resolving task aliases, error out if the alias matches the
name of the current task to prevent infinite recursion:

Example Makefile.toml:

    ## error:
    ##    thread 'main' has overflowed its stack
    ##    fatal runtime error: stack overflow
    ##    Aborted (core dumped)
    [tasks.rec]
    alias = "rec"


sagiegurari:
unfortunately, there are endless ways to get into infinite loop.
wish we could do a more bulletproof solution but I'll accept this one as well.
thanks a lot for the PR. I'll merge it once the CI finishes

phi-gamma:
> unfortunately, there are endless ways to get into infinite loop.

I just realized that as well, see the updated commit.

sagiegurari:
great stuff!!! thanks a lot.

Issue #263, opened 2019-07-11
Profile setting is lost when recursively calling cargo make


Sushisource:
**Describe the bug**
The profile setting is lost when recursively calling cargo make (ex: to go from workspace=false mode to workspace=true)

**To Reproduce**
This script repros the bug in a cargo workspace with at least two members
```
[env]
CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = "true"

[tasks.runme]
workspace = false
dependencies = ["build-native-flow"]
command = "echo"
args = ["${CARGO_MAKE_PROFILE}"]

[tasks.build-native-flow]
workspace = false
category = "Build"
# Recursively call cargo make to go back to using workspace mode, thus running build-native
# for all non-excluded crates in the workspace
command = "cargo"
args = ["make", "--profile=${CARGO_MAKE_PROFILE}", "build-native"]

[tasks.build-native]
description = "Runs the rust compiler."
category = "Build"
command = "echo"
args = ["${CARGO_MAKE_PROFILE}"]
```

Output:
```
(master) λ cargo make --profile production runme                                                                                      (cargomaketest) 16:58:04
[cargo-make] INFO - cargo make 0.21.0
[cargo-make] INFO - Using Build File: Makefile.toml
[cargo-make] INFO - Task: runme
[cargo-make] INFO - Profile: production
[cargo-make] INFO - Running Task: init
[cargo-make] INFO - Running Task: build-native-flow
[cargo-make] INFO - Execute Command: "cargo" "make" "--profile=production" "build-native"
[cargo-make] INFO - cargo make 0.21.0
[cargo-make] INFO - Using Build File: Makefile.toml
[cargo-make] INFO - Task: build-native
[cargo-make] INFO - Profile: production
[cargo-make] INFO - Running Task: init
[cargo-make] INFO - Running Task: workspace
[cargo-make] INFO - cargo make 0.21.0
[cargo-make] INFO - Using Build File: Makefile.toml
[cargo-make] INFO - Task: build-native
[cargo-make] INFO - Profile: development
[cargo-make] INFO - External file not found or is not a file, skipping.
[cargo-make] INFO - Running Task: init
[cargo-make] INFO - Running Task: build-native
[cargo-make] INFO - Execute Command: "echo" "development"
development
[cargo-make] INFO - Running Task: end
[cargo-make] INFO - Build Done  in 0 seconds.
/home/spencer/scratch/cargomaketest
[...]

PR #309, opened 2019-11-09
Pass profiles in workspaces.

daxpedda:
Added a setting to enable passing down profiles to workspace members.

-  Is the name `CARGO_MAKE_PASS_PROFILE` alright?
-  Any improvements to the description in the documentation are welcome.

cc #263.

### DIFF
--- a/docs/_includes/content.md
+++ b/docs/_includes/content.md
@@ -1125,6 +1125,7 @@ In addition to manually setting environment variables, cargo-make will also auto
 * **CARGO_MAKE_PROFILE** - The current profile name in lower case (should not be manually modified by global/task env blocks)
 * **CARGO_MAKE_ADDITIONAL_PROFILES** - The additional profile names in lower case, seperated with a ';' character (should not be manually modified by global/task env blocks)
 * **CARGO_MAKE_CARGO_PROFILE** - The [cargo profile](https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections) name mapped from the **CARGO_MAKE_PROFILE** (unmapped value will default to CARGO_MAKE_PROFILE value)
+* **CARGO_MAKE_PASS_PROFILE** - Pass profile down to workspace member's spawned tasks.
 * **CARGO_MAKE_RUST_VERSION** - The rust version (for example 1.20.0)
 * **CARGO_MAKE_RUST_CHANNEL** - Rust channel (stable, beta, nightly)
 * **CARGO_MAKE_RUST_TARGET_ARCH** - x86, x86_64, arm, etc ... (see rust cfg feature)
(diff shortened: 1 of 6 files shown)


docs/_includes/content.md
```diff
@@ -1125,6 +1125,7 @@ In addition to manually setting environment variables, cargo-make will also auto
 * **CARGO_MAKE_PROFILE** - The current profile name in lower case (should not be manually modified by global/task env blocks)
 * **CARGO_MAKE_ADDITIONAL_PROFILES** - The additional profile names in lower case, seperated with a ';' character (should not be manually modified by global/task env blocks)
 * **CARGO_MAKE_CARGO_PROFILE** - The [cargo profile](https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections) name mapped from the **CARGO_MAKE_PROFILE** (unmapped value will default to CARGO_MAKE_PROFILE value)
+* **CARGO_MAKE_PASS_PROFILE** - Pass profile down to workspace member's spawned tasks.
```
sagiegurari: cargo make doesn't set this one. users do. so i do'nt think it should be in this list. but... i can take care of the docs later if you wish.
if you do want to write about it, try in the workspace support section maybe?

src/lib/execution_plan.rs
```diff
@@ -168,6 +168,12 @@ fn create_workspace_task(crate_info: CrateInfo, task: &str) -> Task {
             make_line.push_str(&log_level);
             make_line.push_str(" --env CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER=");
             make_line.push_str(&member_name);
+
+            if envmnt::is_or("CARGO_MAKE_PASS_PROFILE", false) {
```
sagiegurari: how about
* always pass the profile in the cli - this will require you to update the tests, however anyway i would like a test for this
* if env is true pass the same one as you do now, if false, pass the default profile. this will make things more consistent. I could later on, default the env to true in the internal makefile and publish a backward breaking version.
* can we rename to CARGO_MAKE_USE_WORKSPACE_PROFILE - helps to understand this is workspace related.

sagiegurari:
@daxpedda i want this to be default behavior so created a new branch and deleted old dev branch.
it caused the PR to close. sorry!!!

docs/_includes/content.md
daxpedda: I moved it. Please feel free to give any advice or improve it, I'm not quite sure of the quality.

src/lib/execution_plan.rs
daxpedda: Done.
I really need some guidance on making a test here, I played around a bit, but not really sure how I can make an effective test.

sagiegurari:
merged. thanks a lot.
i'll probably change this to default behavior with ability to turn it off as this fix is really important.

PR #314, opened 2019-11-10
Fix `CARGO_MAKE_TASK_ARGS` not passed down to workspace members.

### LINKED ISSUE #312: `CARGO_MAKE_TASK_ARGS` is not passed down to workspace members.
**Describe the bug**
`CARGO_MAKE_TASK_ARGS` is not passed down to workspace members.

**To Reproduce**
Create a workspace, use some task arguments and `echo ${CARGO_MAKE_TASK_ARGS}` to see that they turn out to be empty.

**Code Sample**
```toml
[tasks.dummy]
script_runner = "@shell"
script = [
	'''
	echo ${CARGO_MAKE_TASK_ARGS}.
	'''
]
```

daxpedda:
Fixes #312.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Migrate to envmnt::expand #308
 * \[**backward compatability break**\] Workspace profile now passed on to members #309 (thanks @daxpedda)
 * New CARGO_MAKE_WORKING_DIRECTORY environment variable #311 (thanks @daxpedda)
-* Fix CARGO_MAKE_TASK_ARGS not being passed down to workspace members #313
+* Fix CARGO_MAKE_TASK_ARGS not being passed down to workspace members #314
 
 ### v0.23.0 (2019-10-16)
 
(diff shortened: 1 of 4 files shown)


sagiegurari:
if cli args are passed, you are ignoring them.
maybe default based on tge env as you did only if no cli args are passed? does it make sense?

daxpedda:
> if cli args are passed, you are ignoring them.

So the reason why I'm ignoring them is because environment variables are passed automatically to workspace members, so if I don't touch the variable, it will just be passed down on it's own.

I assume another solution would be to manually add task args to the member task creation like I did with profiles in `create_workspace_task`. Would that be better?

daxpedda:
Is that better?

daxpedda:
I can't figure out adding a test, sry about that 😞.
Don't have a linux box set up at the moment, so can't play around with the tests locally.

sagiegurari:
I'll check it out later today.
I don't have a Windows machine, so happy to see workspaces work well there :)

sagiegurari:
i'm merging this one in and i'll write a test for it.
thanks a lot for the PR

daxpedda:
Thanks for writing the tests!

PR #318, opened 2019-11-11
Expand environment variables in `tasks.cwd`

### LINKED ISSUE #317: Expand environment variables in tasks.cwd
**Features Description**
Ability to set current working directory in task depending on a environment variable.

Reuse the current `cwd` key in `tasks` to expand variables the unix variable expansion style way that is already used throughout `cargo make`.

**Code Sample**
```toml
[tasks.dummy]
cwd = "${CARGO_MAKE_TESTING}"
command = "an_executable"
```

daxpedda:
This is potentially a breaking change, I marked it as such in the changelog.

Fixes #317.


src/lib/runner.rs
```diff
@@ -38,7 +38,7 @@ where
             if cwd.len() > 0 {
                 let directory = envmnt::get_or("CARGO_MAKE_WORKING_DIRECTORY", "");
 
-                environment::setup_cwd(Some(cwd));
+                environment::setup_cwd(Some(&super::environment::expand_value(cwd)));
```
sagiegurari: it may expand to an empty string so i think another len>0 is needed.
naturally do it before line 39.
also, why super?

sagiegurari:
thanks. this makes a lot of sense.

src/lib/runner.rs
daxpedda: > also, why super?

Im baffled myself 😕.

sagiegurari:
looks great. merging. thanks a lot

PR #331, opened 2019-12-02
Add `CARGO_MAKE_CARGO_HOME`.

### LINKED ISSUE #330: Add `CARGO_MAKE_CARGO_HOME` environment variable
**Features Description**
A variable to get `CARGO_HOME`.
I'm not sure if there are other use cases, but I'm using it in combination with `--remap-path-prefix`.

Probably the easiest method is to use [`home`](https://crates.io/crates/home) to get the correct path and pass it through.

**Code Sample**
```toml
[tasks.build-release]
env = { RUSTFLAGS = "--remap-path-prefix ${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}= --remap-path-prefix ${CARGO_MAKE_CARGO_HOME}=" }
command = "cargo"
args = ["build", "--release"]
```

daxpedda:
I wasn't sure were to add it, so I put it in `setup_cwd`.

Fixes #330.

### DIFF
--- a/src/lib/environment/mod_test.rs
+++ b/src/lib/environment/mod_test.rs
@@ -4,6 +4,7 @@ use crate::types::{ConfigSection, EnvFileInfo, EnvValueUnset, Task};
 use indexmap::IndexMap;
 use std::collections::HashMap;
 use std::env;
+use std::path::Path;
 use std::{thread, time};
 
 #[test]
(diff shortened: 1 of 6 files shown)


daxpedda:
I don't think the CI errors are my fault, right?

sagiegurari:
thanks a lot.
setup_env is the main place to define env vars so maybe move a call to setup_cargo_env (new func) from there.
setup_cwd is not good as its called many times for each task that wants to change the current working directory.

daxpedda:
> setup_cwd is not good as its called many times for each task that wants to change the current working directory.

I think that's correct, because the `HOME_PATH` depends on the working directory, so if we want to change working directory it should also change `HOME_PATH`.

[`cargo_home_with_cwd`](https://docs.rs/home/0.5.1/home/fn.cargo_home_with_cwd.html).

sagiegurari:
ok, got it. thanks

src/lib/environment/mod_test.rs
```diff
@@ -561,6 +562,31 @@ fn setup_env_empty() {
...
+    assert_eq!(
+        envmnt::get_or_panic("CARGO_MAKE_CARGO_HOME"),
+        home::cargo_home().unwrap().to_str().unwrap()
+    );
+}
+
+#[test]
+fn setup_cargo_home_overwrite() {
+    let path = Path::new("path");
+    envmnt::set("CARGO_HOME", path);
```
sagiegurari: can you restore this one to existing/none value? it might be the cause of the ci error (didn't validate, but nice to cleanup such changes)

daxpedda: Done.

daxpedda:
Yeah that fixed it, thanks.

sagiegurari:
thanks a lot

PR #356, opened 2020-01-04
Add target triple.

daxpedda:
This adds a `CARGO_MAKE_RUST_TARGET_TRIPLE` environment variable, mainly because I'm sick of building it over and over by hand again.

Until now my main use case was building with `build-std`, which needs a defined `--target`, but I'm sure there are quite some use cases for this.

This relies on sagiegurari/rust_info#1, I thought this should be upstream, correct me if I should just implement it here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1349,6 +1349,7 @@ In addition to manually setting environment variables, cargo-make will also auto
 * **CARGO_MAKE_RUST_TARGET_OS** - windows, macos, ios, linux, android, etc ... (see rust cfg feature)
 * **CARGO_MAKE_RUST_TARGET_POINTER_WIDTH** - 32, 64
 * **CARGO_MAKE_RUST_TARGET_VENDOR** - apple, pc, unknown
+* **CARGO_MAKE_RUST_TARGET_TRIPLE** - x86_64-unknown-linux-gnu, x86_64-apple-darwin, x86_64-pc-windows-msvc, etc ...
 * **CARGO_MAKE_CRATE_HAS_DEPENDENCIES** - Holds true/false based if there are dependencies defined in the Cargo.toml or not (defined as *false* if no Cargo.toml is found)
 * **CARGO_MAKE_CRATE_IS_WORKSPACE** - Holds true/false based if this is a workspace crate or not (defined even if no Cargo.toml is found)
 * **CARGO_MAKE_CRATE_WORKSPACE_MEMBERS** - Holds list of member paths (defined as empty value if no Cargo.toml is found)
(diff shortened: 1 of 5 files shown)


sagiegurari:
you can define this env var from other env vars in the stable toml without any rust coding.
wouldn't it be better?

daxpedda:
> you can define this env var from other env vars in the stable toml without any rust coding.
> wouldn't it be better?

The thing is `x86_64-apple-darwin` for example is missing the vendor, so this cann't be built properly in stable toml without some form of scripting, which is possible of course.

In any case, I was mainly doing this to reuse it in #358, I'm not aware how to pull environment variables out of stable toml from inside the `cargo-make` crate.

I can change this of course if it's still desired.

src/lib/environment/mod.rs
```diff
@@ -371,6 +371,10 @@ fn setup_env_for_rust() -> RustInfo {
         "CARGO_MAKE_RUST_TARGET_VENDOR",
         &rustinfo.target_vendor.unwrap_or("unknown".to_string()),
     );
+    envmnt::set(
```
sagiegurari: i wonder if maybe just use set_optional here and not default to unknown.
as for triple, unknow is not really useful. what do you think?

README.md
```diff
@@ -1349,6 +1349,7 @@ In addition to manually setting environment variables, cargo-make will also auto
 * **CARGO_MAKE_RUST_TARGET_OS** - windows, macos, ios, linux, android, etc ... (see rust cfg feature)
 * **CARGO_MAKE_RUST_TARGET_POINTER_WIDTH** - 32, 64
 * **CARGO_MAKE_RUST_TARGET_VENDOR** - apple, pc, unknown
+* **CARGO_MAKE_RUST_TARGET_TRIPLE** - x86_64-unknown-linux-gnu, x86_64-apple-darwin, x86_64-pc-windows-msvc, etc ...
```
sagiegurari: the readme is actually generated from content.md (and few others) by running cargo make generate-readme
so doing a change to this file only won't help.
better to add this line to content.md (in docs/_includes) and run cargo make generate-readme

src/lib/environment/mod.rs
daxpedda: Agreed.

README.md
daxpedda: I added it, sadly I can't run `generate-readme` on my machine because I am currently running windows 😞.

sagiegurari:
looks good to me. thanks a lot. will merge after rust info

daxpedda:
I rebased.

sagiegurari:
@daxpedda i'm merging this one. its really great.
thanks a lot!!!
please see my comments on the next PR.

PR #358, opened 2020-01-04
Implement `CARGO_MAKE_CRATE_TARGET_TRIPLE`.

### LINKED ISSUE #357: Get default build target including from `.cargo/Config.toml`
**Features Description**
A way to get default build targets from `.cargo/Config.toml` or fall back to rustc default.

An environment variable like `CARGO_MAKE_CRATE_TARGET_TRIPLE`.

**Code Sample**
```toml
[tasks.wasm-bindgen]
condition = { env = { CARGO_MAKE_CRATE_TARGET_TRIPLE = "wasm32-unknown-unknown" } }
command = "wasm-bindgen"
```

daxpedda:
I am not sure how to implement a test for this, any guidance would be appreciated.
Depends on #356.

This was the reason why I went for #356 and sagiegurari/rust_info#1.

Fixes #357.


src/lib/execution_plan.rs
```diff
@@ -143,6 +145,44 @@ fn update_member_path(member: &str) -> String {
     member_path
 }
 
+pub(crate) fn crate_target_triple(path: impl Into<PathBuf>) -> Option<String> {
```
sagiegurari: i think we can extract this logic to the crate_info? its under the env folder.

src/lib/environment/mod.rs
```diff
@@ -533,6 +533,15 @@ pub(crate) fn setup_cwd(cwd: Option<&str>) {
                 "CARGO_MAKE_CARGO_HOME",
                 &home::cargo_home_with_cwd(directory_path).ok(),
             );
+            envmnt::set(
```
sagiegurari: i would change to set_or_remove so this would only be set if the crate_target_triple returns a value. otherwise we should remove it as i don't think 'unknown' is valid or useful and better just not have it.

README.md
```diff
@@ -1354,6 +1354,7 @@ In addition to manually setting environment variables, cargo-make will also auto
 * **CARGO_MAKE_CRATE_WORKSPACE_MEMBERS** - Holds list of member paths (defined as empty value if no Cargo.toml is found)
 * **CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER** - Holds the name of the current workspace member being built (only if flow started as a workspace level flow)
 * **CARGO_MAKE_CRATE_LOCK_FILE_EXISTS** - Holds true/false based if a Cargo.lock file exists in current working directory (in workspace projects, each member has a different working directory).
+* **CARGO_MAKE_CRATE_TARGET_TRIPLE** - Gets target triple that will be build with by default, respects .cargo/Config.toml's and ${CARGO_HOME}/Config.toml.
```
sagiegurari: can we make this one in the content.md and not here as readme.md is auto generated via cargo make generate-readme.

src/lib/execution_plan.rs
daxpedda: Done.

src/lib/environment/mod.rs
daxpedda: Done.

daxpedda: I also moved the defaulting to the rust default target triple inside of `crate_target_triple`.

README.md
daxpedda: Done, as mentioned in the other PR, I sadly can't run `generate_readme`.

daxpedda:
Working on the test now.

src/lib/environment/mod.rs
daxpedda: Had to revert it again because it was setting it needlessly on every member, breaking all tests in the process.

daxpedda:
I can't actually run the test that I implemented, again because of linux only.
Will check the CI when sagiegurari/rust_info#1 is merged.

src/lib/environment/crateinfo.rs
```diff
@@ -185,3 +187,41 @@ pub(crate) fn load_from(file_path: PathBuf) -> CrateInfo {
...
 }
+
+pub(crate) fn crate_target_triple(path: impl Into<PathBuf>) -> Option<String> {
+    let path = path.into();
+    let mut target_triple = None;
+
+    let config_folders = path
+        .ancestors()
+        .map(|ancestor| ancestor.join(".cargo"))
+        .chain(env::var("CARGO_MAKE_CARGO_HOME").map(PathBuf::from));
```
sagiegurari: what happens if CARGO_MAKE_CARGO_HOME is not defined?

src/lib/environment/mod.rs
```diff
@@ -533,6 +533,11 @@ pub(crate) fn setup_cwd(cwd: Option<&str>) {
                 "CARGO_MAKE_CARGO_HOME",
                 &home::cargo_home_with_cwd(directory_path).ok(),
             );
+            envmnt::set_or_remove(
+                "CARGO_MAKE_CRATE_TARGET_TRIPLE",
+                &crateinfo::crate_target_triple(env::current_dir().unwrap())
```
sagiegurari: not a fan of the unwrap as it may panic. can we change to unwrap_or(".")? would it work?

src/lib/execution_plan.rs
```diff
@@ -198,6 +198,13 @@ fn create_workspace_task(crate_info: CrateInfo, task: &str) -> Task {
             make_line.push_str(" ");
             make_line.push_str(&task);
 
+            if let Some(triple) = environment::crateinfo::crate_target_triple(
+                env::current_dir().unwrap().join(&member_path),
```
sagiegurari: again, this unwrap is a bit concerning. maybe unwrap_or(".")? wdyt?

src/lib/execution_plan.rs
```diff
@@ -198,6 +198,13 @@ fn create_workspace_task(crate_info: CrateInfo, task: &str) -> Task {
...
+            ) {
+                make_line.push_str(" --env CARGO_MAKE_CRATE_TARGET_TRIPLE=");
```
sagiegurari: not sure i understand this one, if env loads this env var during cargo-make initialization anyway... why do we need to pass it to the members? they would anyway load it on their own and set it and ignore the provided value. no?

src/lib/execution_plan_test.rs
```diff
@@ -414,6 +414,49 @@ cd -"#
     assert!(task.env.is_none());
 }
 
+#[test]
+#[cfg(target_os = "linux")]
+fn create_workspace_with_cargo_config() {
```
sagiegurari: you already build a dummy workspace2 for testing which is great.
can we use it also to test crate_target_triple directly? having some test in crate_info_test.rs?
it would also work for all platforms, not just linux which is even better.

src/lib/environment/crateinfo.rs
daxpedda: Nothing.
```rust
.chain({
    let home: Option<String> = env::var("CARGO_MAKE_CARGO_HOME").map(PathBuf::from);
    home.into_iter()
});
```
`chain` takes an iterator, if `Option` is `None` it will just be an empty iterator.

src/lib/environment/mod.rs
daxpedda: I am not sure 😕.
Maybe we should just fail if we can't aquire the current directory?

src/lib/execution_plan.rs
daxpedda: We want every member to re-evaluate their `CARGO_MAKE_CRATE_TARGET_TRIPLE`, because if a `.cargo/config.toml` is in a member but not in the root, it should only evaluate for this member.

I hope I'm not missing something?

src/lib/environment/mod.rs
daxpedda: Alright, I changed it to fail if we can't get the current directory.

src/lib/execution_plan_test.rs
daxpedda: Done.

daxpedda:
Sorry for the delay!

daxpedda:
I'm honestly lost why these tests fail so badly ... any help would be appreciated.

src/lib/execution_plan.rs
sagiegurari: right, agree. you want each member to re-evaluate.
so why does the workspace level has to also evaluate and provide this value to the members on the cli line? if they do it themselves?
maybe you should test it with and without this line and i think you will see same results in the members...

sagiegurari:
tests are failing because you are doing env::set_current_dir to some dir but afterwards you are not reverting to the previous dir.

daxpedda:
> tests are failing because you are doing env::set_current_dir to some dir but afterwards you are not reverting to the previous dir.

Right ... thank you very much.

src/lib/execution_plan.rs
daxpedda: I did test it, it didn't give the same results.
But I agree, it doesn't have to be through the cli line, this doesn't make any sense, I will try to put it somewhere else.

daxpedda: You were right, I think the rebasing issue that I made played a role here in me not getting this righth 😞.
I'm removing this.

daxpedda: Now that I have removed it, I have no clue how to make a proper test for this.
Mainly I don't see how to access the environment variables of members inside of a real workspace test.
Beforehand I was just using the `task.script` to compare the passed values, which isn't possible now.

src/lib/environment/mod.rs
```diff
@@ -534,6 +534,10 @@ pub(crate) fn setup_cwd(cwd: Option<&str>) {
                 "CARGO_MAKE_CARGO_HOME",
                 &home::cargo_home_with_cwd(directory_path).ok(),
             );
+            envmnt::set_or_remove(
+                "CARGO_MAKE_CRATE_TARGET_TRIPLE",
+                &crateinfo::crate_target_triple(None).or_else(|| rust_info::get().target_triple),
```
sagiegurari: rust_info::get() is short but its an expensive call.
i wonder if we can in the mod call get rust info first and pass it to get crate info to enable sharing of data and prevent that expensive call. wdyt?

daxpedda: Sounds perfectly doable, will look into it in a bit.

daxpedda: I solved it with just getting `CARGO_MAKE_RUST_TARGET_TRIPLE`, instead of polluting all these functions with more arguments I thought.

sagiegurari: but that means we are counting on it that it is invoked first. if its not a param, there is no 'contract' to enforce it. it may fail later on quietly if i changed the order for some dumb reason.

daxpedda: Hm ... 🤔
Should I move it away from `setup_cwd` then? I mean is there a more fitting place for it?

daxpedda: I did a little workaround, but it works now.

src/lib/execution_plan.rs
daxpedda: I managed to implement a test with how its now implemented.

daxpedda:
This should be better now.

src/lib/environment/mod.rs
```diff
@@ -445,7 +449,12 @@ fn setup_env_for_project(config: &Config, crate_info: &CrateInfo) {
 }
 
 /// Sets up the env before the tasks execution.
-pub(crate) fn setup_env(cli_args: &CliArgs, config: &Config, task: &str) -> EnvInfo {
+pub(crate) fn setup_env(
+    cli_args: &CliArgs,
+    config: &Config,
+    task: &str,
+    home: Option<PathBuf>,
```
daxpedda: Seeing that `home` won't be used anywhere, I just consumed it.
I can make this a reference if needed.

sagiegurari:
this looks great!!!
thanks a lot for all the really hard work you invested in those PRs.
I will merge this one today (the code looks good to me), finish few of my items as well and hopefully release a new version tomorrow.

daxpedda:
> thanks a lot for all the really hard work you invested in those PRs.

Thank you for maintaing this amazing project with so much care! ❤️

PR #389, opened 2020-03-20
Display recursion-level like `make(1)`

Ma27:
* Added the env-var `CARGO_MAKE_INTERNAL_RECURSION` to keep track
  of recursive calls to `cargo-make`.

* General information like version of `cargo-make` will only be
  displayed at the first call.

* At the first recursion, the level will be shown by the logger.

### DIFF
--- a/src/lib/cli.rs
+++ b/src/lib/cli.rs
@@ -14,6 +14,7 @@ use crate::environment;
 use crate::logger;
 use crate::logger::LoggerOptions;
 use crate::profile;
+use crate::recursion_level;
 use crate::runner;
 use crate::types::{CliArgs, GlobalConfig};
 use crate::version;
@@ -28,13 +29,17 @@ static DEFAULT_TASK_NAME: &str = "default";
 static DEFAULT_OUTPUT_FORMAT: &str = "default";
 
 fn run(cli_args: CliArgs, global_config: &GlobalConfig) {
+    recursion_level::increase_level();
+
     logger::init(&LoggerOptions {
         level: cli_args.log_level.clone(),
         color: !cli_args.disable_color,
     });
 
-    info!("{} {}", &cli_args.command, &VERSION);
-    debug!("Written By {}", &AUTHOR);
+    if recursion_level::is_first_level() {
+        info!("{} {}", &cli_args.command, &VERSION);
+        debug!("Written By {}", &AUTHOR);
+    }
 
     debug!("Cli Args {:#?}", &cli_args);
     debug!("Global Configuration {:#?}", &global_config);
--- a/src/lib/logger.rs
+++ b/src/lib/logger.rs
@@ -8,6 +8,7 @@
 mod logger_test;
 
 use colored::*;
+use crate::recursion_level;
 use envmnt;
 use fern;
 use log::{Level, LevelFilter};
@@ -123,9 +124,12 @@ pub(crate) fn init(options: &LoggerOptions) {
     envmnt::set("CARGO_MAKE_LOG_LEVEL", &level_name_value);
     envmnt::set_bool("CARGO_MAKE_DISABLE_COLOR", !color);
 
+    let level = recursion_level::recursion_level();
+
     let result = fern::Dispatch::new()
         .format(move |out, message, record| {
             let name = env!("CARGO_PKG_NAME");
+
             let record_level = record.level();
 
             if cfg!(test) {
@@ -139,8 +143,8 @@ pub(crate) fn init(options: &LoggerOptions) {
             let record_level_fmt = get_formatted_log_level(&record_level, color);
 
             out.finish(format_args!(
-                "[{}] {} - {}",
-                &name_fmt, &record_level_fmt, &message
+                "[{}] {} {} - {}",
+                &name_fmt, level, &record_level_fmt, &message
             ));
 
             if record_level == Level::Error {
--- a/src/lib/recursion_level.rs
+++ b/src/lib/recursion_level.rs
@@ -1,0 +1,28 @@
+//! # recursion_level
+//!
+//! Utility functions to keep track of the recursion of cargo-make calls.
+//!
+
+#[cfg(test)]
+#[path = "./recursion_level_test.rs"]
+mod recursion_level_test;
+
+use envmnt;
+
+static RECURSION_ENV_VAR_NAME: &str = "CARGO_MAKE_INTERNAL_RECURSION";
+
+pub(crate) fn recursion_level() -> u32 {
+    envmnt::get_or(RECURSION_ENV_VAR_NAME, "0").parse().expect("failed to retrieve env var")
+}
+
+pub(crate) fn is_first_level() -> bool {
+    recursion_level() == 0
+}
+
+pub(crate) fn increase_level() {
+    envmnt::set(
+        RECURSION_ENV_VAR_NAME,
+        if envmnt::exists(RECURSION_ENV_VAR_NAME) { (recursion_level() + 1).to_string() }
+        else { "0".to_string() }
+    )
+}
--- a/src/lib/recursion_level_test.rs
+++ b/src/lib/recursion_level_test.rs
@@ -1,0 +1,23 @@
+use super::*;
+use envmnt;
+
+#[test]
+fn recursion_level_changes() {
+    // backup environment to avoid having conflicts if
+    // the test is run within `cargo-make`.
+    let rec_lvl = envmnt::get_or(RECURSION_ENV_VAR_NAME, "0");
+    envmnt::remove(RECURSION_ENV_VAR_NAME);
+
+    assert!(is_first_level());
+    assert_eq!(recursion_level(), 0);
+
+    increase_level(); // explicitly set to 0
+    increase_level();
+    assert!(!is_first_level());
+    assert_eq!(recursion_level(), 1);
+
+    increase_level();
+    assert_eq!(recursion_level(), 2);
+
+    envmnt::set(RECURSION_ENV_VAR_NAME, rec_lvl);
+}
--- a/src/lib/mod.rs
+++ b/src/lib/mod.rs
@@ -157,6 +157,7 @@ mod io;
 mod legacy;
 mod logger;
 mod profile;
+mod recursion_level;
 mod runner;
 mod scriptengine;
 mod storage;


src/lib/cli.rs
```diff
@@ -111,11 +115,11 @@ fn run(cli_args: CliArgs, global_config: &GlobalConfig) {
 
     let crate_name = envmnt::get_or("CARGO_MAKE_CRATE_NAME", "");
     if crate_name.len() > 0 {
-        info!("Project: {}", &crate_name);
```
sagiegurari: let keep in info.
when working in workspace, it really helps to see this for each member as it tells you that you finished previous one and started this one.

src/lib/cli.rs
```diff
@@ -111,11 +115,11 @@ fn run(cli_args: CliArgs, global_config: &GlobalConfig) {
...
+        debug!("Project: {}", &crate_name);
     }
-    info!("Build File: {}", &build_file);
```
sagiegurari: all these 3 are pretty important as i see it.
i moved other things to verbose, but these 3 are basic.

src/lib/cli.rs
```diff
@@ -33,8 +35,10 @@ fn run(cli_args: CliArgs, global_config: &GlobalConfig) {
         color: !cli_args.disable_color,
     });
 
-    info!("{} {}", &cli_args.command, &VERSION);
-    debug!("Written By {}", &AUTHOR);
+    if env::var("CARGO_MAKE_INTERNAL_RECURSION").unwrap_or_else(|_| "0".to_string()) != "0" {
```
sagiegurari: can we move all the 'internal recursion' logic to a separate module.
i see duplicate code in many areas of this PR and feels like this should be a really small standalone module with
get, is_first and so on...
and since i saw the logger is using this as well, this new module should not have any inner dependency from other cargo-make modules. 
what do you think?

also instead of that complex line, you can do in the module envmnt::get_or("CARGO_MAKE_INTERNAL_RECURSION", "0".to_string())

src/lib/cli.rs
```diff
@@ -33,8 +35,10 @@ fn run(cli_args: CliArgs, global_config: &GlobalConfig) {
...
+        info!("{} {}", &cli_args.command, &VERSION);
```
sagiegurari: these 2 makes a lot of sense to be printed only on the first recurresion. great idea.

src/lib/logger.rs
```diff
@@ -133,15 +134,24 @@ pub(crate) fn init(options: &LoggerOptions) {
     let result = fern::Dispatch::new()
         .format(move |out, message, record| {
             let name = env!("CARGO_PKG_NAME");
+            let level = env::var("CARGO_MAKE_INTERNAL_RECURSION")
```
sagiegurari: lets use the new module. and in that module, lets use envmnt::get_or("CARGO_MAKE_INTERNAL_RECURSION", "0".to_string()) which is much simpler.

src/lib/logger.rs
```diff
@@ -133,15 +134,24 @@ pub(crate) fn init(options: &LoggerOptions) {
...
+            let mut log_prefix = format!("[{}", &name_fmt).to_string();
+            if level != "0" {
+                log_prefix.push_str(&format!(" [{}]", &level));
+            }
+            log_prefix.push_str("]");
+
             out.finish(format_args!(
-                "[{}] {} - {}",
-                &name_fmt, &record_level_fmt, &message
+                "{} {} - {}",
```
sagiegurari: i would change this one back to the original with added numeric one
"[{}]{} {} - {}",
but now that we added {} we can simply add a new binded variable to it that by default is empty string or any value you want like "[number]" or exmaple.
makes the format more understandable.
log prefix can stay the same and when building the formatted one, we can easily see what is in it.

src/lib/runner.rs
```diff
@@ -234,7 +234,9 @@ fn run_task(flow_info: &FlowInfo, step: &Step) {
     };
 
     if validate_condition(&flow_info, &step) {
-        info!("Running Task: {}", &step.name);
+        if &step.name != "empty" {
```
sagiegurari: lets please not do that :)

src/lib/runner.rs
```diff
@@ -517,16 +519,18 @@ pub(crate) fn run(config: Config, task: &str, env_info: EnvInfo, cli_args: &CliA
         run_protected_flow(&flow_info);
     }
 
-    let time_string = match start_time.elapsed() {
-        Ok(elapsed) => {
-            let mut string = "in ".to_string();
-            string.push_str(&elapsed.as_secs().to_string());
-            string.push_str(" seconds");
+    if task.to_string() != "empty" {
```
sagiegurari: same here.... lets not do special handling for empty.

src/lib/command.rs
```diff
@@ -200,6 +201,10 @@ pub(crate) fn run_command(command_string: &str, args: &Option<Vec<String>>, vali
 /// Runs the given task command.
 pub(crate) fn run(step: &Step) {
     let validate = !step.config.should_ignore_errors();
+    let level = env::var("CARGO_MAKE_INTERNAL_RECURSION")
+        .unwrap_or_else(|_| "0".to_string());
+
+    envmnt::set("CARGO_MAKE_INTERNAL_RECURSION", (level.parse::<i64>().unwrap() + 1).to_string());
```
sagiegurari: not sure this is the right place. this runs any command, not just cargo-make calling cargo-make.
so when cargo-make calls cargo build or test, it goes through here. so you are counting lots of stuff...
also if cargo-make is invoked from script, you missed it (in case of workspaces i generate a script per member).
so i think this logic needs to go to the cli.run at the very start, before any printing starts....

sagiegurari:
got an even better idea for removing those 'empty' task logs.
i'm now going to check if a task is even actionable, meaning if it has command/script/run_task/installation scripts/deps/...
if it has anything that affects the flow or runtime, it will be triggered (and printed), if not, its going to be ignored (and not printed in info level, only verbose).
going to push that really soon

sagiegurari:
just pushed #390 which should remove the need to do any handling for 'empty' tasks.
they will not be printed as well.

src/lib/cli.rs
Ma27: I'd skip things like logging the cargo-make version on recursive calls then: when you have multiple Makefiles in a project and get this message many times, it makes it IMHO harder to actually parse the output in your head.

Ma27:
As you provided an even better solution for most of the concerns I raised (and already published a release :heart_eyes: ), the only thing left here is the tracking of recursive calls.

I moved the logic as requested into its own crate (with a simple test) and reworded PR and commit-message accordingly.

src/lib/runner.rs
Ma27: agreed - your solution is way better :D

src/lib/logger.rs
sagiegurari: still prefer the change to fit my original formatting request.

src/lib/recursion_level.rs
```diff
@@ -0,0 +1,28 @@
...
+pub(crate) fn is_first_level() -> bool {
+    recursion_level() == 0
+}
+
+pub(crate) fn increase_level() {
+    envmnt::set(RECURSION_ENV_VAR_NAME, (recursion_level() + 1).to_string())
+}
+
+pub(crate) fn decrease_level() {
+     envmnt::set(RECURSION_ENV_VAR_NAME, (recursion_level() - 1).to_string())
```
sagiegurari: lets make sure we don't end up <0 here... ya it would mean a bug, but still lets just make sure.

src/lib/recursion_level_test.rs
```diff
@@ -0,0 +1,34 @@
+use super::*;
+use envmnt;
+use indexmap::IndexMap;
+
+#[test]
+fn recursion_level_changes() {
+    // backup environment to avoid having conflicts if
+    // the test is run within `cargo-make`.
+    let all = envmnt::vars();
```
sagiegurari: we can backup just the recursion level, not entire env.

src/lib/recursion_level_test.rs
```diff
@@ -0,0 +1,34 @@
...
+    let keys = all.iter().map(|x| &x.0).collect::<Vec<_>>();
+
+    envmnt::remove_all(&keys);
```
sagiegurari: lest just remove the recursion level env.

src/lib/recursion_level_test.rs
```diff
@@ -0,0 +1,34 @@
...
+    decrease_level();
+    decrease_level();
+    assert!(is_first_level());
+
+    let mut next = IndexMap::new();
+    for (key, value) in all.iter() {
+        next.insert(key.to_string(), value.to_string());
+    }
+
+    envmnt::set_all(&next);
```
sagiegurari: and of course, lets revert only the recursion level.

src/lib/command.rs
```diff
@@ -197,6 +198,8 @@ pub(crate) fn run_command(command_string: &str, args: &Option<Vec<String>>, vali
 pub(crate) fn run(step: &Step) {
     let validate = !step.config.should_ignore_errors();
 
+    recursion_level::increase_level();
```
sagiegurari: you are increasing it when invoking a task, not when invoking a child cargo-make.
so i'm not 100% sure this is the right way.
there are several places where cargo-make calls itself

1. watch task
2. protected flow
3. fork
(1-3 go through create_proxy_task)
4. workspace - so cargo-make creates a script that calls each member.
5. user created tasks - that can be in the command or script.

so as i see it, the best way to ensure this works bulletproof for all scenarios is to increment it in the cli.rs run function right from the beggining. and if not defined, set to 0.
wdyt?

src/lib/cli.rs
```diff
@@ -135,6 +140,8 @@ fn run(cli_args: CliArgs, global_config: &GlobalConfig) {
     } else {
         runner::run(config, &task, env_info, &cli_args);
     }
+
+    recursion_level::decrease_level();
```
Ma27: How would you take care of panics here? `panic::catch_unwind` doesn't seem enough (at least in the test suite).

sagiegurari: it doesn't matter if it panics and there is no need to decrease level.
child processes can't change parent process env.
so you just increment. when it ends, parent is unaffected.

Ma27: Oof, you're absolutely right, of course! Still had the mindset of the previous solution where a decrease was necessary, sorry!

Will fix this tomorrow accordingly :)

src/lib/logger.rs
```diff
@@ -139,8 +142,8 @@ pub(crate) fn init(options: &LoggerOptions) {
             let record_level_fmt = get_formatted_log_level(&record_level, color);
 
             out.finish(format_args!(
-                "[{}] {} - {}",
-                &name_fmt, &record_level_fmt, &message
+                "[{}] {} {} - {}",
+                &name_fmt, recursion_level::recursion_level(), &record_level_fmt, &message
```
Ma27: That's actually a bug since now the recursion level will be determined when writing a log-entry (which can be after the increase e.g. when displaying the version).

Ma27: fixed in the latest push.

src/lib/command.rs
Ma27: should be fixed now.

src/lib/cli.rs
```diff
@@ -47,6 +50,8 @@ fn run(cli_args: CliArgs, global_config: &GlobalConfig) {
         version::check();
     }
 
+    recursion_level::increase_level();
```
sagiegurari: i think this needs to be set at first line of the run. before any invocation of the logger.
otherwise it means you have 2 different levels at same cargo-make process...

src/lib/logger.rs
```diff
@@ -8,9 +8,11 @@
 mod logger_test;
 
 use colored::*;
+use crate::recursion_level;
 use envmnt;
 use fern;
 use log::{Level, LevelFilter};
+use std::env;
```
sagiegurari: where is this used?

src/lib/recursion_level.rs
```diff
@@ -0,0 +1,24 @@
...
+pub(crate) fn recursion_level() -> u32 {
+    envmnt::get_or(RECURSION_ENV_VAR_NAME, "0").parse().expect("failed to retrieve env var")
+}
+
+pub(crate) fn is_first_level() -> bool {
+    recursion_level() == 0
+}
+
+pub(crate) fn increase_level() {
+    envmnt::set(RECURSION_ENV_VAR_NAME, (recursion_level() + 1).to_string())
```
sagiegurari: in order to ensure the first cargo-make stays at 0.
i think this function (and tell me if this makes sense), should only increment if value exists. if not, it should set it to 0 and next invocation of it, will increase it to 1.
it would also resolve your issue with the cli.rs that you delayed setting it.

sagiegurari:
This looks great.
i'm going to merge this and probably do some minor things, but its great.
thanks a lot!!

PR #404, opened 2020-04-02
Fix toolchain check on Windows

Osspial:
The toolchain check was previously using the `true` command to verify that the given toolchain existed, but the `true` command doesn't exist on Windows. This replaces that command with `rustc`.


sagiegurari:
thank you very much

PR #444, opened 2020-08-07
Fix tasks.do-copy-apidocs

haraldh:
1. It didn't work with a non-default target specified e.g. in
   `.cargo/config`
2. It didn't work with `CARGO_MAKE_CRATE_WORKSPACE_MEMBERS`
3. It didn't work when running `cargo make docs-flow` multiple times

### DIFF
--- a/src/lib/descriptor/makefiles/rust.toml
+++ b/src/lib/descriptor/makefiles/rust.toml
@@ -733,17 +733,37 @@ run_task = "do-copy-apidocs"
 description = "Copies the generated documentation to the docs/api directory."
 category = "Documentation"
 private = true
-script_runner = "@shell"
+script_runner = "@duckscript"
 script = [
 '''
-SOURCE_DIRECTORY=${CARGO_MAKE_DOCS_ROOT_FOLDER}/target/doc
-TARGET_DIRECTORY=${CARGO_MAKE_DOCS_ROOT_FOLDER}/docs/api/${CARGO_MAKE_DOCS_SUB_FOLDER}
+if is_empty "${CARGO_MAKE_DOCS_SUB_FOLDER}"
+    SRC_DIR= set ${CARGO_MAKE_DOCS_ROOT_FOLDER}
+else
+    SRC_DIR= set ${CARGO_MAKE_DOCS_ROOT_FOLDER}/${CARGO_MAKE_DOCS_SUB_FOLDER}
+end
 
-echo Source Directory: ${SOURCE_DIRECTORY}
-echo Target Directory: ${TARGET_DIRECTORY}
+if is_dir ${SRC_DIR}/target/${CARGO_MAKE_CRATE_TARGET_TRIPLE}/doc
+    SRC_DIR= set ${SRC_DIR}/target/${CARGO_MAKE_CRATE_TARGET_TRIPLE}/doc
+else
+    SRC_DIR= set ${SRC_DIR}/target/doc
+end
 
-mkdir -p ${TARGET_DIRECTORY}
-mv ${SOURCE_DIRECTORY}/* ${TARGET_DIRECTORY}
+if is_empty "${CARGO_MAKE_DOCS_SUB_FOLDER}"
+    DST_DIR= set ${CARGO_MAKE_DOCS_ROOT_FOLDER}/docs/api
+else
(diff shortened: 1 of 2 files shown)


sagiegurari:
thanks a lot for the PR.
few comments: you are not using the doc root folder env var. any reason for that?
what if a project would like to change it?
second, this won't work on windows, so if you want to add conditions and such, we should convert it to duckscript maybe.

haraldh:
with `CARGO_MAKE_CRATE_WORKSPACE_MEMBERS` the `CARGO_MAKE_DOCS_ROOT_FOLDER` does not point to the current crate where the docs where build when doing `cargo docs-flow`

sagiegurari:
you mean when the project is a workspace? in that case, i guess the best fix is to set it up correctly for those type of projects as well. 
if scripting is a must, than it can be via duckscript because I'm not sure how else it would be cross platform

haraldh:
Ok, here is my take on it with duckscript.

sagiegurari:
Thanks a lot. looks promising. i'm going to locally test it as well. probably tomorrow.

src/lib/descriptor/makefiles/rust.toml
```diff
@@ -733,17 +733,40 @@ run_task = "do-copy-apidocs"
...
 category = "Documentation"
 private = true
-script_runner = "@shell"
+script_runner = "@duckscript"
 script = [
 '''
-SOURCE_DIRECTORY=${CARGO_MAKE_DOCS_ROOT_FOLDER}/target/doc
-TARGET_DIRECTORY=${CARGO_MAKE_DOCS_ROOT_FOLDER}/docs/api/${CARGO_MAKE_DOCS_SUB_FOLDER}
+if is_empty ${CARGO_MAKE_DOCS_SUB_FOLDER}
+    SOURCE_DIRECTORY= set ${CARGO_MAKE_DOCS_ROOT_FOLDER}
```
sagiegurari, line 740: since SOURCE_DIRECTORY is now actually a variable, can we name it source_dir or something short?
same goes for other variables.

src/lib/descriptor/makefiles/rust.toml
```diff
@@ -733,17 +733,40 @@ run_task = "do-copy-apidocs"
...
+else
+    TARGET_DIRECTORY= set ${CARGO_MAKE_DOCS_ROOT_FOLDER}/docs/api/${CARGO_MAKE_DOCS_SUB_FOLDER}
+end
 
 echo Source Directory: ${SOURCE_DIRECTORY}
 echo Target Directory: ${TARGET_DIRECTORY}
 
-mkdir -p ${TARGET_DIRECTORY}
-mv ${SOURCE_DIRECTORY}/* ${TARGET_DIRECTORY}
+if is_dir ${TARGET_DIRECTORY}
```
sagiegurari, line 760: no need to check if the dir exists. you can just remove. it won't break if it doesn't exist.

src/lib/descriptor/makefiles/rust.toml
```diff
@@ -733,17 +733,40 @@ run_task = "do-copy-apidocs"
...
+    SOURCE_DIRECTORY= set ${CARGO_MAKE_DOCS_ROOT_FOLDER}/${CARGO_MAKE_DOCS_SUB_FOLDER}
+end
+
+if is_dir ${SOURCE_DIRECTORY}/target/${CARGO_MAKE_CRATE_TARGET_TRIPLE}/doc
+    SOURCE_DIRECTORY= set ${SOURCE_DIRECTORY}/target/${CARGO_MAKE_CRATE_TARGET_TRIPLE}/doc
+else
+    SOURCE_DIRECTORY= set ${SOURCE_DIRECTORY}/target/doc
+end
+
+if is_empty ${CARGO_MAKE_DOCS_SUB_FOLDER}
```
sagiegurari, line 751: just in case to make sure CARGO_MAKE_DOCS_SUB_FOLDER is defined, let change to:
if is_empty "${CARGO_MAKE_DOCS_SUB_FOLDER}"

haraldh: yes

src/lib/descriptor/makefiles/rust.toml
haraldh: ok

src/lib/descriptor/makefiles/rust.toml
haraldh: ok

sagiegurari:
thanks a lot for the PR. merging

PR #455, opened 2020-09-02
Allow for explicit workspace flow in sub flows

epontan:
Commit fa929b116f78 ("Fix workspace detection for sub flows") caused a
regression when having a root workspace with tasks that relies on each
member to define CARGO_MAKE_CARGO_BUILD_TEST_FLAGS with unique set of
features.

Determine if workspace flow is explicitly set in the task when planning
the execution and take this setting into account when running in a sub
flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## CHANGELOG
 
+### v0.32.5
+
+* Allow for explicit workspace flow in sub flows
+
 ### v0.32.4 (2020-08-28)
 
 * Prevent multiple calls to cleanup task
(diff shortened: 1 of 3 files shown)


sagiegurari:
@epontan thanks a lot for the PR, it looks good.
I rebased it on top a new dev branch. not sure why the build failed, but lets see if it passes now.

epontan:
@sagiegurari thank you! It looks like all passed except the code coverage this time. I'm not sure why it fails. If I run coverage-tarpaulin locally the coverage is increased by half a point.

sagiegurari:
ya, its due to rustfmt compilation broken on nightly. just pushed a workaround, let me restart the GH actions and see how it goes.

sagiegurari:
@epontan can you pull 032.5 changes to your local branch? this will contain the rustfmt workaround so we can see coverage is ok

epontan:
@sagiegurari that seemed to have done the trick! Thanks :+1:

sagiegurari:
@epontan this is really great fix. thanks a lot!!!
merging

PR #458, opened 2020-09-22
Add toolchain to the namespace configuration.

calavera:
This allows setting the toolchain version for alltasks in a common place, and still override it in individual tasks.

It's very useful to manage the toolchain for all tasks across a project. Freeing people from having to keep their local toolchains in sync with other environments, like CI.

This is an example of how I'm using it inside a large project:

```
[config]
    default_to_workspace = false
    toolchain = "nightly-2020-09-20"
```

This PR is missing tests and docs, so I'm creating it as a draft.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/src/lib/runner.rs
+++ b/src/lib/runner.rs
@@ -22,6 +22,7 @@ use crate::logger;
 use crate::profile;
 use crate::scriptengine;
 use crate::time_summary;
+use crate::toolchain;
 use crate::types::{
     CliArgs, Config, DeprecationInfo, EnvInfo, EnvValue, ExecutionPlan, FlowInfo, FlowState,
     RunTaskInfo, RunTaskName, RunTaskRoutingInfo, Step, Task, TaskWatchOptions,
@@ -353,12 +354,9 @@ pub(crate) fn run_task(flow_info: &FlowInfo, flow_state: &mut FlowState, step: &
             // modify step using env and functions
             let mut updated_step = functions::run(&step);
             updated_step = environment::expand_env(&updated_step);
+            toolchain::expand(flow_info, &mut updated_step);
 
-            if updated_step.config.toolchain.is_none()
-                && flow_info.config.config.toolchain.is_some()
-            {
-                updated_step.config.toolchain = flow_info.config.config.toolchain.clone();
-            }
+            debug!("Updated step definition {:#?}", &updated_step.config);
 
             let watch = should_watch(&step.config);
 
--- a/src/lib/runner.rs
+++ b/src/lib/runner.rs
@@ -354,6 +354,12 @@ pub(crate) fn run_task(flow_info: &FlowInfo, flow_state: &mut FlowState, step: &
             let mut updated_step = functions::run(&step);
             updated_step = environment::expand_env(&updated_step);
 
+            if updated_step.config.toolchain.is_none()
+                && flow_info.config.config.toolchain.is_some()
+            {
+                updated_step.config.toolchain = flow_info.config.config.toolchain.clone();
+            }
+
             let watch = should_watch(&step.config);
 
             if watch {
--- a/src/lib/toolchain.rs
+++ b/src/lib/toolchain.rs
@@ -7,7 +7,7 @@
 #[path = "./toolchain_test.rs"]
 mod toolchain_test;
 
-use crate::types::CommandSpec;
+use crate::types::{CommandSpec, FlowInfo, Step};
 use std::process::{Command, Stdio};
 
 #[cfg(test)]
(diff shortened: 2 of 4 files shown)


sagiegurari:
@calavera thanks a lot for the PR. really cool idea.
Your implementation basically adds it only for the main task and maybe tasks invoked by run_task (didn't check).
but it doesn't add it to all tasks.
for example if you have a task as a dependency of another task, i think it will skip your code and the dependency will be invoked with the default toolchain.

so.... it feels to me, like this code should reside in the runner.rs and not execution plan rs.
that's a catch all place (the run_task function).

wdty?

calavera:
> so.... it feels to me, like this code should reside in the runner.rs and not execution plan rs.
> that's a catch all place (the run_task function).
> 
> wdty?

that sounds like a great idea. To be honest, I basically grepped by a place to put the code, and that's the first one I found :rofl: Thanks for pointing me in the right direction.

sagiegurari:
> grepped by a place to put the code, and that's the first one I found

ok, that helps me understand how PRs are born. lol

src/lib/runner.rs
```diff
@@ -354,6 +354,12 @@ pub(crate) fn run_task(flow_info: &FlowInfo, flow_state: &mut FlowState, step: &
             let mut updated_step = functions::run(&step);
             updated_step = environment::expand_env(&updated_step);
 
+            if updated_step.config.toolchain.is_none()
+                && flow_info.config.config.toolchain.is_some()
+            {
+                updated_step.config.toolchain = flow_info.config.config.toolchain.clone();
+            }
+
```
sagiegurari, line 362: if you look 2 lines up, you can see that i call external modules from this function to update the task.
so functions updates the task and than environment updates the task.
maybe we can expose in toolchain module, also a function to update the task.
this way, this function doesn't become too much of a bloat (it is already), and the logic is in the specific modules.
wdyt?

sagiegurari (changes requested):

src/lib/runner.rs
calavera: that makes sense. I think this won't apply to all tasks yet, unfortunately, tasks that define a script are not wrapped with the toolchain change, right?

sagiegurari: oh... you raise a good point here. this whole PR might be a bit problematic as it will add the toolchain to all tasks including tasks that just do copy file.
all commands will become 'rustup sometoolchain original command'
so i'll get 'rustup stable cp file1 file2'

that is an issue.
so it might not be a practical feature.

maybe we want to limit it only to cargo commands? are we missing something in this case?
i don't care about scripts. scripts are there for a specific purpose, but if someone writes cargo test in a script instead of command and args, ya, they will loose capabilities (like they lose today that auto installation capability)

calavera: > if someone writes cargo test in a script instead of command and args, ya, they will loose capabilities (like they lose today that auto installation capability)

this is exactly how I discovered it :rofl:

src/lib/runner.rs
```diff
@@ -353,6 +354,9 @@ pub(crate) fn run_task(flow_info: &FlowInfo, flow_state: &mut FlowState, step: &
             // modify step using env and functions
             let mut updated_step = functions::run(&step);
             updated_step = environment::expand_env(&updated_step);
+            toolchain::expand(flow_info, &mut updated_step);
```
sagiegurari, line 357: can we call this modify or something?
expand is something you do with env.

src/lib/toolchain.rs
```diff
@@ -66,3 +66,10 @@ fn has_toolchain(toolchain: &str) -> bool {
         .expect("Failed to check rustup toolchain")
         .success()
 }
+
+pub(crate) fn expand(flow_info: &FlowInfo, step: &mut Step) {
+    println!("{:?}", flow_info.config.config.toolchain);
```
sagiegurari, line 71: can we drop this?

src/lib/toolchain.rs
```diff
@@ -66,3 +66,10 @@ fn has_toolchain(toolchain: &str) -> bool {
...
+    if step.config.toolchain.is_none() && flow_info.config.config.toolchain.is_some() {
```
sagiegurari, line 72: can we do it only if this task as a command value and that command is 'cargo'?

src/lib/types.rs
```diff
@@ -1698,6 +1698,8 @@ pub struct ConfigSection {
     pub windows_load_script: Option<Vec<String>>,
     /// acts like load_script if runtime OS is Mac (takes precedence over load_script)
     pub mac_load_script: Option<Vec<String>>,
+    /// Toolchain to use for all tasks, this can be overwritten by individual tasks
+    pub toolchain: Option<String>,
```
sagiegurari, line 1702: lets rename to default_toolchain

sagiegurari (changes requested):
overall looks good.
got few comments and we need docs+tests.

calavera:
This is starting to look really not what I wanted. It'll be very hard to cover all corner cases, so I'm going to close it.

sagiegurari:
that's ok. happens to me as well.
thanks a lot for the effort and idea.

PR #459, opened 2020-09-26
Fix for env_files loading

wolf4ood:
Hi,
i found out that when loading `env_files` with profiles , if a profile is not matched the loading process is interrupted and the eventual other files in the list are skipped.

In order to show this behaviour  i've created the test `set_env_files_for_config_profile_inverse ` by copying `set_env_files_for_config_profile` inverting the positions of the two files in the test.

I'm not sure if it's a wanted behaviour or a bug. In case of a bug in this PR there is the fix

Thanks :)


sagiegurari:
@wolf4ood thanks a lot for the fix. its a definite bug :)
merging

PR #463, opened 2020-10-01
Provide more detailed error message when env-script fails

epontan:
There is limited information given in the error output when an
env-script return a non-zero exit code, which can be quite confusing as
there is no context of where it failed (unless verbose output is turned
on).

Avoid confusion by providing more context about the environment when the
script fails.

BEFORE:
 > cargo make
 [cargo-make] INFO - cargo make 0.32.5
 [cargo-make] ERROR - Error while executing command, exit code: 128
 [cargo-make] WARN - Build Failed.

AFTER:
 > cargo make
 [cargo-make] INFO - cargo make 0.32.5
 [cargo-make] ERROR - Error while evaluating script for env: GIT_ROOT, exit code: 128
 script: [
     "git rev-parse --show-toplevel",
 ]
 stdout: ""
 stderr: "fatal: not a git repository (or any of the parent directories): .git
 "
 [cargo-make] WARN - Build Failed.

### DIFF
--- a/src/lib/environment/mod.rs
+++ b/src/lib/environment/mod.rs
@@ -31,40 +31,47 @@ use rust_info::types::{RustChannel, RustInfo};
 use std::env;
 use std::path::{Path, PathBuf};
 
-fn evaluate_env_value(env_value: &EnvValueScript) -> String {
+fn evaluate_env_value(key: &str, env_value: &EnvValueScript) -> String {
     match command::run_script_get_output(&env_value.script, None, &vec![], true, Some(false)) {
         Ok(output) => {
             let exit_code = output.0;
             let stdout = output.1;
+            let stderr = output.2;
 
-            command::validate_exit_code(exit_code);
+            if exit_code != 0 {
+                error!(
+                    concat!(
+                        "Error while evaluating script for env: {}, exit code: {}\n",
+                        "script:\n{:#?}\n",
+                        "stdout:\n{}\n",
+                        "stderr:\n{}\n"
+                    ),
+                    key, exit_code, env_value.script, &stdout, &stderr
+                );
+            }
 
             debug!("Env script stdout:\n{}", &stdout);
 
-            if exit_code == 0 {
-                let multi_line = match env_value.multi_line {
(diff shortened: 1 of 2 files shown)


sagiegurari:
That's really great. Let me check the code out.

src/lib/environment/mod.rs
```diff
@@ -31,13 +31,24 @@ use rust_info::types::{RustChannel, RustInfo};
...
-            command::validate_exit_code(exit_code);
+            let stderr = output.2;
+
+            if exit_code != 0 {
+                error!(
+                    concat!(
+                        "Error while evaluating script for env: {}, exit code: {}\n",
+                        "script: {:#?}\n",
+                        "stdout: \"{}\"\n",
+                        "stderr: \"{}\""
```
sagiegurari, line 47: can we drop the wrapper " here? its a bit confusing in the output.
we can do
stdout:\n{}\n
stderr:\n{}\n
instead. wdyt?

src/lib/environment/mod.rs
```diff
@@ -31,13 +31,24 @@ use rust_info::types::{RustChannel, RustInfo};
...
             let stdout = output.1;
-
-            command::validate_exit_code(exit_code);
+            let stderr = output.2;
+
+            if exit_code != 0 {
+                error!(
+                    concat!(
+                        "Error while evaluating script for env: {}, exit code: {}\n",
+                        "script: {:#?}\n",
```
sagiegurari, line 45: can we change this one to
script:\n{:#?}\n

sagiegurari (commented):
looks great. can we play with the output formatting a bit?

src/lib/environment/mod.rs
epontan: Sure, I couldn't really decide one ideal output format anyway.

src/lib/environment/mod.rs
epontan: Sure, np.

sagiegurari:
Looks great @epontan 
thanks a lot for the PR

PR #467, opened 2020-10-02
ISSUE_TEMPLATE: Change formatting

Kreyren:
I believe that this formatting makes the issues more readable.

Signed-off-by: Jacob Hrbek <kreyren@rixotstudio.cz>

PR #468, opened 2020-10-02
CI: Attempt for optimization

Kreyren:
This is submitted to avoid CI trigger on issues that are not changing the source code alike https://github.com/sagiegurari/cargo-make/pull/467

Signed-off-by: Jacob Hrbek <kreyren@rixotstudio.cz>


sagiegurari:
thanks, but this adds complexity that doesn't give much value, but makes the yml less readable.
i also don't think i want to run only for src/* and cargo.toml, for exmaple, what if i change the ci.yml (like in this PR)?
so, i think i prefer to keep it simple as it is today.
but thanks a lot for thinking about the improvement.

PR #498, opened 2020-11-30
Add ability to use 'object' form in dependencies to define task path

### LINKED ISSUE #497: Propose syntax-sugar for referring to tasks in other crates
I have a workspace with many crates. As part of the packaging process, I'd like to copy some files (html templates) between them. This copy stage depends on some preprocessing defined in `src/web/Makefile.toml`. In the root `Makefile.toml` I am having to create a dummy task that changes directory and runs the target I want.

**src/web/Makefile.toml**
```toml
[tasks.yarn]
command = "yarn"

[tasks.css]
command = "yarn"
args = ["build"]
dependencies = ["yarn"]

[tasks.dist]
command = "cp"
args = ["-a", "static/.", "dist/"]
dependencies = ["css"]
```

**Makefile.toml**
```toml
[tasks.web-dist]
cwd = "src/web"
command = "cargo"
args = ["make", "dist"]

[tasks.web-release]
cwd = "src"
command = "cp"
args = ["-r", "web/dist", "application/dist"]
dependencies = ["web-dist"]
```

The combined result of which is:
- compile the css
- copy it into dist along with the static files
- copy the files into the application crate

My question: there is a little boilerplate task here, would it be an idea to provide some sugar and allow me to define a dependency as follows instead?:

```toml

[tasks.web-release]
cwd = "src"
command = "cp"
args = ["-r", "web/dist", "application/dist"]
dependencies = ["web:dist"]
```

web:dist could either be a path or a crate in the workspace, and it refers to the dist task inside the Makefile there.

arlyon:
From the changes in the readme: 

> Tasks may also depend on tasks in other files. To do this, specify the
dependency with the object format, providing the path. Cargo make
will use this path as it would any other supplied on the command line:
if a filename is supplied, it searches that file. Otherwise it uses
the default `Makefile.toml`.
> 
>```toml
>[tasks.install]
>command = "mv"
>args = ["src/B/out", "src/C/static"]
>dependencies = [
>  { name = "compile", path = "src/B" },
>  { name = "clean", path = "src/C/tasks.toml" },
>]
>```

I originally used `cwd` instead of `path` but being able to select a file as the `cwd` didn't feel right. Happy to squash commits if needed. 

Closes #497


arlyon:
I'll add some coverage tests after feedback :)

sagiegurari:
thanks a lot!!! I'll do a deeper dive tomorrow but in general i like the approach and the toml format you decided on.
but, I'll take a closer look tomorrow.

src/lib/execution_plan.rs
```diff
@@ -323,13 +324,31 @@ fn is_workspace_flow(
...
     allow_private: bool,
 ) {
-    let task_config = get_normalized_task(config, task, true);
+    if let Some(path) = &task.path {
+        // this is refering to a task in another file
+        // so we create a proxy task to invoke it
+        let proxy_name = format!("{}_proxy", task.name);
+        let mut proxy_task = create_proxy_task(&proxy_name, true, false);
+
+        proxy_task.command = Some("cargo".to_string());
```
sagiegurari, line 339: why do we need command and args? those are set by the create_proxy_task already which also populates all the needed args, not just task name.

src/lib/runner.rs
```diff
@@ -496,7 +496,11 @@ fn create_watch_task(task: &str, options: Option<TaskWatchOptions>) -> Task {
     task_config
 }
 
-fn create_proxy_task(task: &str, allow_private: bool, skip_init_end_tasks: bool) -> Task {
+pub(crate) fn create_proxy_task(
```
sagiegurari, line 499: runner is using execution_plan and now execution_plan is using runner.
this is cyclic dependency which i prefer to break if possible.
maybe we can move this function to another module? proxy.rs (probably bad name....)?

src/lib/types.rs
```diff
@@ -979,6 +979,85 @@ pub struct Task {
     pub mac: Option<PlatformOverrideTask>,
 }
 
+/// A dependency, defined either as a string or as a Dependency object
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[serde(untagged)]
+pub enum StringTaskIdentifier {
```
sagiegurari, line 985: this is used for dependencies, maybe just call it DepedencyIdentifier?
regardless, i wouldn't write String in the name.

src/lib/types.rs
```diff
@@ -979,6 +979,85 @@ pub struct Task {
...
+    /// A full dependency definion (potentially in a different file)
+    Definition(TaskIdentifier),
+    /// A string dependency definition (its name in the current file)
+    Name(String),
+}
+
+impl From<&str> for StringTaskIdentifier {
+    fn from(s: &str) -> Self {
```
sagiegurari, line 993: can we avoid single character variable names? i know it might only be me :)

src/lib/types.rs
```diff
@@ -979,6 +979,85 @@ pub struct Task {
...
+        StringTaskIdentifier::Name(s.to_string())
+    }
+}
+
+impl StringTaskIdentifier {
+    /// Get the name of a dependency
+    pub fn name(&self) -> &str {
+        match self {
+            StringTaskIdentifier::Definition(d) => &d.name,
+            StringTaskIdentifier::Name(s) => s,
```
sagiegurari, line 1003: same, d and s :)

src/lib/types.rs
```diff
@@ -979,6 +979,85 @@ pub struct Task {
...
+
+    /// Adorn the TaskIdentifier with a namespace
+    pub fn with_namespace(self, namespace: &str) -> Self {
+        match self {
+            StringTaskIdentifier::Definition(mut d) => {
+                d.name = get_namespaced_task_name(namespace, &d.name);
+                StringTaskIdentifier::Definition(d)
+            }
+            StringTaskIdentifier::Name(s) => {
+                StringTaskIdentifier::Name(get_namespaced_task_name(namespace, &s))
```
sagiegurari, line 1015: same d and s

src/lib/types.rs
```diff
@@ -979,6 +979,85 @@ pub struct Task {
...
+    pub name: String,
+    /// The path to the makefile the task resides in
+    pub path: Option<String>,
+}
+
+impl std::fmt::Display for TaskIdentifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match &self.path {
+            Some(p) => write!(f, "{}:{}", p, self.name),
+            None => write!(f, "{}", self.name),
```
sagiegurari, line 1034: same p and f

src/lib/types.rs
```diff
@@ -979,6 +979,85 @@ pub struct Task {
...
+        }
+    }
+}
+
+impl Into<TaskIdentifier> for StringTaskIdentifier {
+    fn into(self) -> TaskIdentifier {
+        match self {
+            StringTaskIdentifier::Definition(d) => d,
+            StringTaskIdentifier::Name(s) => TaskIdentifier {
+                name: s,
```
sagiegurari, line 1054: same d and s

src/lib/types.rs
```diff
@@ -1091,15 +1170,13 @@ impl Task {
...
-                        for task in &dependencies {
-                            modified_dependencies.push(get_namespaced_task_name(namespace, &task));
-                        }
-
-                        self.dependencies = Some(modified_dependencies);
+                    if let Some(dependencies) = &self.dependencies {
+                        self.dependencies = Some(
+                            dependencies
+                                .iter()
+                                .map(|d| d.to_owned().with_namespace(namespace))
```
sagiegurari, line 1177: d :)

sagiegurari (changes requested):
amazing work. really love it. 
also liked the toml definition.
just small nits and than setup some tests and we are good to go.

src/lib/types.rs
arlyon: I usually only use them for lambdas and patterns but I agree they're more useful when you have rust_analyzer holding your hand! I'll get those fixed :)

src/lib/proxy_task.rs
```diff
@@ -0,0 +1,57 @@
+use std::env;
+
+use crate::{logger, profile, types::Task};
+
+pub(crate) fn create_proxy_task(
```
sagiegurari, line 5: you should also move all create_proxy tests from runner_test to new proxy_task_test file

sagiegurari (commented):
looks amazing, lets just move the tests from runner_test to proxy_task_test for the create_proxy_task fn

src/lib/proxy_task.rs
arlyon: Yes, of course haha I knew there was _something_. All fixed.

sagiegurari:
@arlyon i think the wrong imports there are causing a build error.
i think you just need to import super

arlyon:
> @arlyon i think the wrong imports there are causing a build error.
> i think you just need to import super

Should be fixed now, but I wasn't getting those errors locally. Thanks for the patience :)

sagiegurari:
@arlyon i'm merging this one. thanks a lot for the great help.
i'll probably release a new version soon.

PR #508, opened 2021-01-20
Fix link to profile section

132ikl:


sagiegurari:
@132ikl thanks a lot!!!
this file is auto generated from:
https://github.com/sagiegurari/cargo-make/blob/master/docs/_includes/content.md
so would it be possible to fix it there as well?

132ikl:
@sagiegurari Done.

PR #531, opened 2021-03-10
Minor change: Use "args" where there is more than one "arg"

gemmaro:
Replaced [`arg`](https://doc.rust-lang.org/std/process/struct.Command.html#method.arg)s with [`args`](https://doc.rust-lang.org/std/process/struct.Command.html#method.args).
This commit is not a bug fix.

### DIFF
--- a/src/lib/version.rs
+++ b/src/lib/version.rs
@@ -32,8 +32,7 @@ fn get_version_from_output(line: &str) -> Option<String> {
 
 fn get_latest_version() -> Option<String> {
     let result = Command::new("cargo")
-        .arg("search")
-        .arg("cargo-make")
+        .args(&["search", "cargo-make"])
         .output();
 
     match result {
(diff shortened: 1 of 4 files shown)


src/lib/version.rs
```diff
@@ -32,8 +32,7 @@ fn get_version_from_output(line: &str) -> Option<String> {
 
 fn get_latest_version() -> Option<String> {
     let result = Command::new("cargo")
-        .arg("search")
-        .arg("cargo-make")
+        .args(&["search", "cargo-make"])
```
sagiegurari, line 35: not sure this is more readable.... lets drop this one.

sagiegurari:
nice. is that a new function? don't remember it when i first developed cargo-make.

sagiegurari:
merging. thanks a lot for this quality change :)

PR #535, opened 2021-03-29
Fix makers formatting

### LINKED ISSUE #533: Broken makers output formatting in terminal (Windows)
### Describe The Bug
Broken `makers` output formatting in terminal. Probably only on Windows (git bash, ps, cmd)

image

### To Reproduce
Run `makers` on Windows.

### How to resolve it

We were debugging it in [Trunk](https://trunkrs.dev/) and [this code](https://github.com/thedodd/trunk/blob/7b3cdde99daafd5bccdf7dab5bc742a3bba2a5bb/src/main.rs#L20-L23) (+ related dependency) should be enough to resolve it:
```rust
#[cfg(windows)]
if let Err(err) = ansi_term::enable_ansi_support() {
    eprintln!("error enabling ANSI support: {:?}", err);
}
```
_Note:_ `cargo make` works well, I assume `cargo` calls similar command by itself.

---

OT: Hi @sagiegurari. I'm just integrating `cargo-make` to https://github.com/MoonZoon/MoonZoon - `cargo make` is still the best task runner 🙂

MartinKavik:
Fixes #533 

There are two commits, the first fixes the issue, the second one fixes `cargo make ci-flow` on Windows.

The reason why formatting works when you run `makers` indirectly (through `cargo run` or `cargo make`) is that `cargo` use a special print function: https://github.com/rust-lang/cargo/blob/e4aebf0a039ca3787f3ce98a9d469a3187f83706/src/cargo/core/shell.rs#L327-L340 with Windows-only code. _Note:_ There are multiple Unix/Windows "hacks" in the linked file to make the shell work as expected.

Another way is to enable ANSI support globally, just like Trunk or other apps do it. This approach has been used in this PR.  We need to call a native API to enable it: https://github.com/ogham/rust-ansi-term/blob/ff7eba98d55ad609c7fcc8c7bb0859b37c7545cc/src/windows.rs

### DIFF
--- a/src/lib/mod.rs
+++ b/src/lib/mod.rs
@@ -195,5 +195,9 @@ mod version;
 
 /// Handles the command line arguments and executes the runner.
 pub fn run_cli(command_name: String, sub_command: bool) {
+    #[cfg(windows)]
+    if let Err(err) = ansi_term::enable_ansi_support() {
+        eprintln!("error enabling ANSI support: {:?}", err);
+    }
     cli::run_cli(command_name, sub_command)
 }
--- a/src/lib/mod.rs
+++ b/src/lib/mod.rs
@@ -158,6 +158,10 @@
 //! [Apache 2](https://github.com/sagiegurari/cargo-make/blob/master/LICENSE) open source license.
 //!
 
+// Dependencies used in the binary `makers`:
+#[cfg(windows)]
+use ansi_term as _;
+
 #[macro_use]
 extern crate log;
 #[macro_use]
@@ -195,9 +199,5 @@ mod version;
 
 /// Handles the command line arguments and executes the runner.
 pub fn run_cli(command_name: String, sub_command: bool) {
-    #[cfg(windows)]
-    if let Err(err) = ansi_term::enable_ansi_support() {
-        eprintln!("error enabling ANSI support: {:?}", err);
-    }
     cli::run_cli(command_name, sub_command)
 }
--- a/src/makers.rs
+++ b/src/makers.rs
@@ -13,7 +13,7 @@
 //!
 
 #[cfg(test)]
-#[path = "./makers_test.rs"]
+#[path = "makers_test.rs"]
 mod makers_test;
 
 fn get_name() -> String {
--- a/src/makers.rs
+++ b/src/makers.rs
@@ -21,6 +21,8 @@ fn get_name() -> String {
 }
 
 fn main() {
+    #[cfg(windows)]
+    let _ = ansi_term::enable_ansi_support();
     let name = get_name();
     cli::run_cli(name, false);
 }
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -22,3 +22,8 @@ cargo make ci-flow
 ````
 
 * There are many automatic unit tests as part of the library which provide full coverage of the functionality.<br>Any fix/enhancement must come with a set of tests to ensure it's working well.
+
+- _For Windows users_: You need to allow to run PowerShell scripts (see [About Execution Policies](https:/go.microsoft.com/fwlink/?LinkID=135170) for more info). Run in PS:
+    ```ps
+    Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope CurrentUser
+    ```
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,9 @@ serde_ignored = "^0.1"
 shell2batch = "^0.4.2"
 toml = "^0.5"
 
+[target.'cfg(windows)'.dependencies]
+ansi_term = "0.12.1"
+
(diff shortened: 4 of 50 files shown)


src/lib/mod.rs
```diff
@@ -195,5 +195,9 @@ mod version;
 
 /// Handles the command line arguments and executes the runner.
 pub fn run_cli(command_name: String, sub_command: bool) {
+    #[cfg(windows)]
+    if let Err(err) = ansi_term::enable_ansi_support() {
+        eprintln!("error enabling ANSI support: {:?}", err);
```
sagiegurari, line 200: not sure we need to print it to the users.... if it doesnt work, no big deal i guess. wdyt?
also, why not do it in makers.rs? since cargo is already fixing it for us, we can just push the fix to the makers executable only.

sagiegurari:
@MartinKavik thanks a lot!!!
few questions:

>  second one fixes cargo make ci-flow on Windows.

what does this mean? i'm running ci flow on windows today, so what was i missing?
why do we need all those removals of './'?
(again, i don't have windows, so i only look at the github windows execution which seems ok.

> _For Windows users_: You need to allow to run PowerShell scripts

can you explain this one? why do we need powershell for cargo make? i usually default to cmd.

src/lib/mod.rs
MartinKavik: I wrote it in `makers.rs` first but then I've moved it to CLI to satisfy linters (otherwise it breaks the rule `not used crate ansi_term` or something like that). 
I've added `eprintln` for easier user reporting when it fails.

So
I'll move it back to `makers.rs` and satisfy/disable the linter rule for it somehow.
Should I remove that `eprintln` call?

sagiegurari: i wonder how to silence that linter for specific case, so in future i don't forget some crates in...
maybe than, just add some bool from makers/main -> cli that tell it if to use it or not... crappy hack.

eprintln -> ya, lets remove it. i'm not sure users would find that helpful and many people already complain that cargo make prints out too much stuff.

MartinKavik:
> what does this mean? i'm running ci flow on windows today, so what was i missing?
why do we need all those removals of './'?

image

It's pretty "standard error" for many tools on Windows. The first related issue I was able to find: https://github.com/rust-lang/rustfmt/issues/1873. It happens at least in Git Bash and PowerShell. Maybe it will be fixed in the next `rustfmt` version or it already works in nightly or there are other factors, who knows :) I would be probably able to track the real problem, but I don't want to kill time unless necessary.

---

> can you explain this one? why do we need powershell for cargo make? i usually default to cmd.

`ci-flow` with the default PowerShell settings:

image

One test:

image

You can switch between `cmd` and `poweshell` multiple ways and a Windows update has changed the default shell to PS. I don't know which settings affect the cargo-make scripts but probably there is a way.

src/lib/mod.rs
MartinKavik: Code updated in the latest commit.

From https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-crate-dependencies:
> This lint is "allow" by default because it can provide false positives depending on how the build system is configured. For example, when using Cargo, a "package" consists of multiple crates (such as a library and a binary), but the dependencies are defined for the package as a whole. If there is a dependency that is only used in the binary, but not the library, then the lint will be incorrectly issued in the library.

src/makers.rs
```diff
@@ -21,6 +21,8 @@ fn get_name() -> String {
 }
 
 fn main() {
+    #[cfg(windows)]
+    let _ = ansi_term::enable_ansi_support();
```
sagiegurari, line 25: question, do we need the let _ here?

sagiegurari:
@MartinKavik thanks a lot for the answers :) 

* ./ issue - feels like rustfmt issue. strange one by the way. but thanks for explaining and let's keep your fix.
* powershell - forgot about that test. i usually run that in CI which has powershell scripting enabled, so never saw it.
I'll probably extend your comment in the contribution doc a bit.

Will probably merge it later today/tomorrow.
I'll first create a new dev branch for it.

src/makers.rs
MartinKavik: Alternative would be something like:
```
#[allow(unused_must_use)]
{
      #[cfg(windows)]
      ansi_term::enable_ansi_support();
}
```
I didn't test it but I've found this workaround on stackoverflow or a Rust forum / issue. 

This doesn't work for some reasons:
```rust
#[allow(unused_must_use)]
#[cfg(windows)]
ansi_term::enable_ansi_support();
```

sagiegurari: lets leave it. thanks for the explanation.

sagiegurari:
Ran few manual tests on linux. looks awesome. thanks again @MartinKavik 
you can now update the second PR so i can merge it as well (i tested it also).

PR #547, opened 2021-04-10
Support custom crate target triple

ark0f:


sagiegurari:
@ark0f really cool, thanks a lot!

PR #548, opened 2021-04-10
Add `CARGO_MAKE_CRATE_TARGET_DIR` env variable

ark0f:


src/lib/environment/crateinfo.rs
```diff
@@ -187,48 +186,62 @@ pub(crate) fn load_from(file_path: PathBuf) -> CrateInfo {
...
+    let path = env::current_dir().ok()?;
 
-        for config_folder in config_folders {
-            let config_file = config_folder.join("config");
+    let config_file = path
+        .ancestors()
+        .map(|ancestor| ancestor.join(".cargo"))
+        .chain(home)
+        .map(|config_file| config_file.join("config"))
+        .filter_map(|config_file| {
```
sagiegurari, line 209: i'm guessing the regression might be here, but didn't investigate.

sagiegurari (changes requested):
@ark0f thanks a lot for the PR. this env var is actually super useful.
i see a test is failing so maybe worth to check what is the regression there.

sagiegurari:
@ark0f also i merged your other PR, so you need to resolve merge conflicts

sagiegurari:
@ark0f looks awesome. lets wait for the CI to finish and than i'll merge it.
thanks a lot!

sagiegurari (approved):

ark0f:
All tests pass after I added `#[ignore]` to `get_crate_target_dir` test (like previous contributor did with `get_crate_target_triple`), but I still don't know why changing CWD can affect tests

sagiegurari:
> I still don't know why changing CWD can affect tests

running cargo test will run tests in parallel so changing cwd will break other tests that look at the file system.
when you add ignore to a test, i postpone it to later and run it in a single thread run (cargo make splits and runs its tests in 2 batches, parallel and single threaded).

ark0f:
Oops, forgot about import

sagiegurari:
looks great. thanks again. merging

PR #554, opened 2021-04-29
Respect `build.target` when getting target dir

### LINKED ISSUE #557: CARGO_MAKE_CRATE_TARGET_DIRECTORY does not respect workspace
### Describe The Bug
If a crate is in a workspace, CARGO_MAKE_CRATE_TARGET_DIRECTORY will still be in the current directory instead of the root of the workspace

### To Reproduce
- Create workspace directory and add project to it:
```sh
$ mkdir workspace
$ cd workspace
$ cargo new project
```
- Create virtual manifest for workspace, adding project to it
```sh
$ cat << EOF > Cargo.toml
[workspace]
members = ["project"]
EOF
```
- Add Makefile to print CARGO_MAKE_CRATE_TARGET_DIRECTORY (remove the additional `\` before `$` if you're editing the file manually)
```sh
$ cd project
$ cat << EOF > Makefile.toml
[env]
TARGET_DIR = "abc"

[tasks.print-target-dir]
script = '''
echo Target Dir: \${CARGO_MAKE_CRATE_TARGET_DIRECTORY}
'''
EOF

```
- Run `print-target-dir` task for project
```
$ cargo make print-target-dir
```
### Output
```console
$ cargo make print-target-dir
[cargo-make] INFO - cargo make 0.32.17
[cargo-make] INFO - Project: project
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: print-target-dir
[cargo-make] INFO - Running Task: print-target-dir
Target Dir: target
[cargo-make] INFO - Build Done in 0.03 seconds.
```

Found target: (equivalent to) `/path/to/workspace/project/target`
Expected target: `/path/to/workspace/target`

The expected target can be found by running `cargo metadata`, in key `target_directory`:

```
$ cargo metadata --format-version 1 --no-deps
```

### (Possible) Solution

I would personally suggest using [cargo metadata](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html) to get the target directory, instead of trying to guess it by manually looking at the environment.

ark0f:
* Use `cargo_metadata` crate
* Respect `build.target` when getting target dir
* Introduce `RustTarget` to cope better with better changes
* Add `CARGO_MAKE_CRATE_CUSTOM_TRIPLE_TARGET_DIRECTORY` env var

Fixes #557

### DIFF
--- a/src/lib/environment/crateinfo.rs
+++ b/src/lib/environment/crateinfo.rs
@@ -11,9 +11,9 @@ use crate::command;
 use crate::types::{CargoMetadata, CrateDependency, CrateInfo};
 use fsio;
 use glob::glob;
-use serde::{Deserialize, Deserializer};
 use serde_json;
 use std::env;
+use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
@@ -195,18 +195,30 @@ struct CargoConfig {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 struct CargoConfigBuild {
-    #[serde(default, deserialize_with = "deserialize_target")]
-    target: Option<String>,
+    target: Option<RustTarget>,
     target_dir: Option<PathBuf>,
 }
 
-fn deserialize_target<'de, D: Deserializer<'de>>(
-    deserializer: D,
-) -> Result<Option<String>, D::Error> {
-    Option::<String>::deserialize(deserializer)?
-        .map(|target| target.trim_end_matches(".json").to_string())
-        .map(Ok)
-        .transpose()
+#[derive(Debug, Deserialize)]
+#[serde(from = "PathBuf")]
+struct RustTarget(PathBuf);
+
+impl RustTarget {
+    fn name(&self) -> &str {
+        self.0.file_stem().unwrap().to_str().unwrap()
+    }
+}
+
+impl From<PathBuf> for RustTarget {
+    fn from(buf: PathBuf) -> Self {
+        Self(buf)
+    }
+}
+
+impl AsRef<OsStr> for RustTarget {
+    fn as_ref(&self) -> &OsStr {
+        self.0.as_ref()
+    }
 }
 
 fn get_cargo_config(home: Option<PathBuf>) -> Option<CargoConfig> {
@@ -241,6 +253,7 @@ pub(crate) fn crate_target_triple(
     get_cargo_config(home)
         .and_then(|config| config.build)
         .and_then(|build| build.target)
+        .map(|target| target.name().to_string())
         .or(default_target_triple)
 }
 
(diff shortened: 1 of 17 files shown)


ark0f:
cargo FS structure looks like this:
```
target
\---debug
\---x86_64-custom-target
    \---debug
```
so it can be wrong behavior. Is it better to introduce a new env var?

src/lib/environment/crateinfo.rs
```diff
@@ -194,8 +195,18 @@ struct CargoConfig {
...
     target: Option<String>,
-    target_dir: Option<String>,
+    target_dir: Option<PathBuf>,
+}
+
+fn deserialize_target<'de, D: Deserializer<'de>>(
+    deserializer: D,
+) -> Result<Option<String>, D::Error> {
+    Option::<String>::deserialize(deserializer)?
+        .map(|target| target.trim_end_matches(".json").to_string())
```
sagiegurari, line 207: why do we need it in the deserialize? why not keep it simple and do it in the usage as today?

sagiegurari:
Thanks a lot for the continued effort on this one :)

> Is it better to introduce a new env var?

can you explain a bit? what new var did you have in mind? how would they be used if there were 2, meaning how do users know which one to use?

src/lib/environment/crateinfo.rs
ark0f: 'Cause it will lead us to code repetition. We need to trim string in `crate_target_dir` and `crate_target_triple`, but original string can be useful for env var like `CARGO_MAKE_IS_CUSTOM_TARGET_TRIPLE`

ark0f:
> can you explain a bit?

Behavior with custom target triple: 
* `cargo` creates `target/debug` folder with build scripts in it
* `cargo` creates `target/my_custom_triple` with the rest

I don't know can `cargo` store there something else and will behavior be changed in the future.

What about workspaces? `cargo` ignores `.cargo/config` of member if you call `cargo` in workspace root, but don't if you do it in member's folder, so you see that member will store its artifacts in `target/my_custom_triple` when others do it in `target`

> what new var did you have in mind?

Something like `CARGO_MAKE_CRATE_CUSTOM_TRIPLE_TARGET_DIRECTORY`

> how would they be used if there were 2, meaning how do users know which one to use?

User will decide which directory he needs - folder of custom target triple (`target/my_custom_triple`) or folder of host (`target`)

sagiegurari:
@ark0f thanks a lot for the answers and PR. i agree with your approach of adding another env var.
tell me when its ready :)

sagiegurari:
looks awesome @ark0f thanks a lot

@pheki can you please verify that your use case is fixed with this PR?

After that i'll merge this into the 0.33 branch

pheki:
Oh, I missed this PR... It's great and I've verified, it fixes #557.

Curiously, I'm also using a custom target, but its being set on `[env]` then passed in `cargo build --target=${RUST_TARGET}`, so I won't be able to use this new env var, but the current approach of joining the path works fine.

sagiegurari:
@pheki thanks for confirming :)

@ark0f merging this. thanks a lot!!!

Issue #552, opened 2021-04-28
coverage-flow raise errors for binary projects with strict CLI arguments handling


elonca:
Hi,

First, thank you for developing such a good tool!

### Describe The Bug

I got an issue with the coverage-flow task for projects producing binaries for which unknown CLI arguments raise errors.
During the coverage-flow task, an error appears just like if the binary was invoked with wrong arguments.

### To Reproduce

Cargo.toml:

```toml
[package]
name = "cargo-make-cov-bug"
version = "0.1.0"
authors = ["emmanuel lonca"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
clap = "2.33.3"
```

main.rs:

```rust
use clap::App;

fn main() {
    App::new("My Super Program").get_matches();
}

```

First, execute `cargo make` (no issue at this point):

```
[cargo-make] INFO - cargo make 0.32.17
[cargo-make] INFO - Project: cargo-make-cov-bug
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: default
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Execute Command: "cargo" "fmt"
[cargo-make] INFO - Execute Command: "cargo" "build" "--all-features"
   Compiling libc v0.2.94
   Compiling bitflags v1.2.1
   Compiling unicode-width v0.1.8
   Compiling strsim v0.8.0
   Compiling ansi_term v0.11.0
   Compiling vec_map v0.8.2
   Compiling textwrap v0.11.0
   Compiling atty v0.2.14
   Compiling clap v2.33.3
   Compiling cargo-make-cov-bug v0.1.0 (/home/lonca/projects/cargo-make-cov-bug)
    Finished dev [unoptimized + debuginfo] target(s) in 5.60s
[cargo-make] INFO - Execute Command: "cargo" "test" "--all-features"
   Compiling cargo-make-cov-bug v0.1.0 (/home/lonca/projects/cargo-make-cov-bug)
    Finished test [unoptimized + debuginfo] target(s) in 0.44s
     Running target/debug/deps/cargo_make_cov_bug-3ae0a44675a0669e

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

[cargo-make] INFO - Build Done in 6.54 seconds.
```

Then, `cargo make coverage-flow`:

```
[...]

PR #555, opened 2021-04-29
kcov should now only be run on test executables

elonca:
During the coverage-flow, kcov is run on both test executables and project binaries (see [#552](https://github.com/sagiegurari/cargo-make/issues/552)).

The following changes keep the same initial set of binaries from the `debug/deps` directory, but removes from it the ones for which a copy exist in the `debug` directory. This way, the project binaries should now not be given to kcov.

The check for copies relies on [GNU cmp](https://en.wikipedia.org/wiki/Cmp_(Unix)) which should be installed on all recent Linux and MacOS systems.


sagiegurari:
thanks a lot for the PR. very interesting solution there. i'm kicking in the CI to see that it works well.

sagiegurari:
@elonca thanks a lot for the PR, seems to work well :)
merging

elonca:
Good news !
Glad to help :)

PR #575, opened 2021-07-30
Reformat toml files with taplo-cli 0.4.0

jayvdb:


sagiegurari:
ran it myself with latest formatter version. i think we can skip this PR.

PR #593, opened 2021-09-26
Remove requirement of "nightly" channel for clippy and format ci flows

damszew:
**Problem I wanted to solve:**
I want to use `ci-flow` with clippy and formating check on my CI but I'm using only "stable" channel in my workflows.

**Reasons:**
`clippy` and `rustfmt` don't require "nightly" channel for some time now

Maybe we can drop that condition?

### DIFF
--- a/src/lib/descriptor/makefiles/rust.toml
+++ b/src/lib/descriptor/makefiles/rust.toml
@@ -237,13 +237,7 @@ dependencies = ["pre-check-format", "check-format", "post-check-format"]
 [tasks.check-format-ci-flow]
 description = "Runs cargo fmt --check if conditions are met."
 category = "Test"
-condition = { env_set = [
-  "CARGO_MAKE_RUN_CHECK_FORMAT",
-], channels = [
-  "nightly",
-], platforms = [
-  "linux",
-] }
+condition = { env_set = ["CARGO_MAKE_RUN_CHECK_FORMAT"], platforms = ["linux"] }
 run_task = "check-format-flow"
 
 [tasks.format-flow]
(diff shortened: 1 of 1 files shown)


sagiegurari:
there is a task to run clippy differntly for stable/beta. i think you might have missed that

damszew:
Did you mean `cargo make clippy`? But it would require adding a separate job to my gh actions workflow. Not to mention that it won't be as clean as using `ci-flow`.

sagiegurari:
ok i get you. so for ci flow, i decided to run those only on nightly.
reason is that, if formatting is messed up, its messed up regardless of your rust channel and since we all run in stable, beta and nightly, lets pick nightly with latest stuff to see that it works well and speed up stable and beta by not running it there as well.
same goes for clippy.

however, if you check check-format and clippy tasks, they do not enforce that directly so they can be used on all rust channels.

if you want to run it on stable and beta as well, just override that task in  your makefile and remove the condition.

damszew:
I did the workaround you mentioned and it is ok but...

It seems kinda wrong. Condition there was prepared so it's easy to turn this flow on or off via env var - but now I need to overwrite it to turn it on.

Why not drop "nightly" requirement and use env variables in workflow to choose if clippy or format should be run?

sagiegurari:
i explained why. basically it is to save time on the ci. no need to run check formatting on stable beta and nightly is the results are the same.
built and test yes naturally. formatting is a waste of time to run on all 3.
this is how i view it, but as you can see, its easily modifiable by anyone via external makefile.

damszew:
What I've meant is that it could be controlled in other means than checking "nightly" channel, like setting up env vars.

Nevertheless, I'll close this PR as I've got workaround

PR #595, opened 2021-10-06
Implement detailed toolchain specifier

### LINKED ISSUE #594: Let toolchain describe a minimal supported rust version
### Feature Description
```
[tasks.rustc-version-stable]
toolchain = "1.56" # required because we're using edition 2021
```

Unfortunately, by the time 1.57 stabilizes, the above will suddenly complain to devs using the latest stable release.

Support a version check for the toolchain. Multiple syntaxes come to mind:
```
[tasks.rustc-version-stable]
toolchain = ">= 1.56"
toolchain = "stable >= 1.56"
toolchain = { channel = "stable", version = ">= 1.56" }
toolchain = { channel = "stable", feature = ["edition_2021"] } # optional, but most specific and declarative
```

The first two could perhaps be a short version of the second.

The last is probably too much to ask, but could use names from specific RFCs. Just don't know how to get rustc to tell me which ones are supported, and maintaining that list here is not such a good idea?

---

- [x] I am willing to implement this in a PR

It seems parsing the output of `rustc --version` is already done in `rust_info`, so perhaps it could be a good idea to reuse that and put it into the public interface there?

https://github.com/sagiegurari/cargo-make/blob/ee4e8b40319532079750d26c7d2415c0c6fbc306/src/lib/toolchain.rs#L60-L68

would then parse the output of `--version` additionally to checking succeeding and do the version comparison.

---

For a real world example, consider https://github.com/yewstack/yew/blob/master/packages/yew-macro/Makefile.toml#L1-L5 which sets `toolchain = "1.51"`, to enforce support of `const_generics`. Changing this to `toolchain = ">= 1.51"` would run into less problems.

WorldSEnder:
As described in #594, this let's us specify a minimal version expected to be present for toolchains with descriptive names like `stable` so that users that didn't (or couldn't) yet update their rust version can error instead of running into compile errors later due to missing features (like `edition_2021`).

Closes #594 

- [x] Implement MVI
- [ ] Add test cases
- [ ] Add documentation

### DIFF
--- a/src/lib/command_test.rs
+++ b/src/lib/command_test.rs
@@ -110,7 +110,7 @@ fn run_command_for_toolchain() {
         let mut task = Task::new();
         task.command = Some("echo".to_string());
         task.args = Some(vec!["test".to_string()]);
-        task.toolchain = Some(toolchain.to_string());
+        task.toolchain = Some(toolchain.into());
 
         let step = Step {
             name: "test".to_string(),
--- a/src/lib/installer/cargo_plugin_installer.rs
+++ b/src/lib/installer/cargo_plugin_installer.rs
@@ -10,10 +10,11 @@ mod cargo_plugin_installer_test;
 use crate::command;
 use crate::installer::crate_version_check;
 use crate::toolchain::wrap_command;
+use crate::types::ToolchainSpecifier;
 use envmnt;
 use std::process::Command;
 
-fn is_crate_installed(toolchain: &Option<String>, crate_name: &str) -> bool {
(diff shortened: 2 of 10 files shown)


sagiegurari:
ya, if nightly fails the coverage drops as lots of tests are nightly only.

sagiegurari:
@WorldSEnder the code looks great.
I'll just wait for the CI to finish and i'll merge it and release it.
thanks a lot for all the work you did and feel free to raise (and implement) more issues :)

PR #603, opened 2021-10-28
List category

grbd:
Hi,
I've recently added an option that allows you to list the tasks based on the category.
so in addition to --list-all-steps there's also now
```
--list-category-steps <CATEGORY>
```

Part of the reason I added this in is that I'm working on a embedded rust project
and wanted to be able to show just the tasks associated with building / flashing with a simple command.
So the idea being I can add in all my build / flash tasks under a single common ESP32 category
then have an alias in my Makefile.Toml such as
```
[tasks.list]
description = "List all of the associated build / flash tasks - ESP32"
command = "cargo"
args = ["make", "--list-category-steps", "ESP32"]

[tasks.flashid]
description = "Get the flashid / details from the esp32"
command = "esptool"
args = ["--chip", "esp32", "-p", "${ESP32_FLASH_SERIAL}", "flash_id", "${@}"]
category = "ESP32"
```


src/lib/cli_commands/list_steps.rs
```diff
@@ -12,8 +13,13 @@ use crate::io;
 use crate::types::{Config, DeprecationInfo};
 use std::collections::BTreeMap;
 
-pub(crate) fn run(config: &Config, output_format: &str, output_file: &Option<String>) -> u32 {
-    let (output, count) = create_list(&config, output_format);
+pub(crate) fn run(
+    config: &Config,
+    output_format: &str,
+    output_file: &Option<String>,
+    category: &str,
```
sagiegurari, line 20: maybe this should be Option? instead of hardcoded 'default'?

src/lib/cli_commands/list_steps.rs
```diff
@@ -51,13 +61,17 @@ pub(crate) fn create_list(config: &Config, output_format: &str) -> (String, u32)
...
 
         if !is_private {
-            count = count + 1;
-
             let category = match task.category {
                 Some(value) => value,
                 None => "No Category".to_string(),
             };
 
+            if category_filter != "default" && category != category_filter {
```
sagiegurari, line 69: we should use an option instead of this. otherwise you can't filter by category = default

src/lib/cli_commands/list_steps.rs
```diff
@@ -99,6 +113,10 @@ pub(crate) fn create_list(config: &Config, output_format: &str) -> (String, u32)
 
     let post_key = if markdown { "**" } else { "" };
     for (category, tasks) in &categories {
+        if category_filter != "default" && category != category_filter {
```
sagiegurari, line 116: same

src/lib/cli_commands/list_steps_test.rs
```diff
@@ -18,7 +18,7 @@ fn run_empty() {
         tasks,
     };
 
-    let count = run(&config, "default", &None);
+    let count = run(&config, "default", &None, "default");
```
sagiegurari, line 21: and for tests, thiswould be None

src/lib/types.rs
```diff
@@ -110,6 +110,8 @@ pub struct CliArgs {
     pub print_only: bool,
     /// List all known steps
     pub list_all_steps: bool,
+    /// List steps for a given category
+    pub list_category_steps: String,
```
sagiegurari, line 114: Option<String>

src/lib/types.rs
```diff
@@ -146,6 +148,7 @@ impl CliArgs {
             skip_tasks_pattern: None,
             print_only: false,
             list_all_steps: false,
+            list_category_steps: "default".to_string(),
```
sagiegurari, line 151: default should be None

sagiegurari (changes requested):
Overall looks good.
I think we should not use the default string hardcoded and instead do option<string> instead otherwise 'default' category won't work and it would be better design.

sagiegurari:
@grbd just want to make sure you saw the comments in this PR

grbd:
Yep spotted them, will have a look later
Just been finishing up my esp32 setup

sagiegurari:
@grbd friendly ping about this one as well :)

grbd:
Ok I think I've fixed this now, let me know if it still needs to change
I'm still a bit new at the way rust handles ownership and pointers.

sagiegurari:
> For my own project I've actually moved to just outputing a list of things via the coloured duckscript
so I'll add an example for that later on via a separate pull request.

That actually sounds like a really cool solution. better than this one.
would love to see it.

sagiegurari:
@grbd thanks for the PR, merging.
i'm going to do few changes but overall looks great.

PR #613, opened 2021-12-01
Examples for showing a list of tasks via duckscript / rust-script colored text

grbd:
Hi,
I've attached a couple of examples of using colored text ether via duckscript (which is the better quicker option)
or via rust-script to show a list of tasks.

It's mostly a copy / paste of an embedded project I'm working on so you may want to reformat it perhaps.
I've found using duckscript functions to define the color formatting the better way of doing things.

The only thing I couldn't figure out how to suppress were the info messages before / after the output.

### DIFF
--- a/examples/usage-duckscript/Makefile.toml
+++ b/examples/usage-duckscript/Makefile.toml
@@ -1,0 +1,10 @@
+# This is the top level cargo make file
+
+extend = [
+
+  # Include the usage toml file
+  # This is an example of using colored text via duckscript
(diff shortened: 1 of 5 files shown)


sagiegurari:
thanks @grbd didn't think it would be a hardcoded list. i think i'll pass on that one for now.
maybe if we have a solution to show all tasks with filtering/sorting/.... it might be something that we can incoporate.
for now i think i'll go with your other PR

PR #624, opened 2022-01-12
Add Cargo features `edition2021`

Magicloud:
This is required when setting `package.edition` to `2021`.


Magicloud:
My bad. This is removed now.

PR #638, opened 2022-03-10
Update regex

Dylan-DPC:
Update regex to `1.5.5` as per [security advisory](https://blog.rust-lang.org/2022/03/08/cve-2022-24713.html)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ home = "^0.5"
 indexmap = { version = "^1", features = ["serde-1"] }
 ignore = "^0.4"
 log = "^0.4"
-regex = "^1"
+regex = "1.5.5"
 run_script = "^0.9"
 rust_info = "^0.3.1"
 semver = "^1.0"


sagiegurari:
@Dylan-DPC I think there might be some misunderstanding of how semver works.
* your changes - limit regex  1.5.5<=crate<1.6
* current setup - 1<=crate<2

this basically means, than if i push your change and regex go up to 1.6 I will not have this change. Since regex is at version >=1 that would include fixes but no breaking changes.

Also cargo install ALWAYS takes the highest value possible based on dependencies.  so if you do cargo-install today (without your changes) 1.5.5 would be used anyway.

so unless i'm missing something, this change doesn't add anything and will not impact new users doing cargo install but will limit them in the future once regex goes above 1.5

feel free to reopen if I missed anything and this is needed. for now, based on my understanding, it is not.
and thanks a lot for caring and trying to ensure we do not have any security issue in cargo-make. really appreciate it :)

Dylan-DPC:
>     your changes - limit regex 1.5.5<=crate<1.6,  current setup - 1<=crate<2

no, this is false. My change does `^1.5.5` which is  `1.5.5 <= x < 2.0.0` 
([docs on semver compatibility](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) 

> Also cargo install ALWAYS takes the highest value possible based on dependencies. so if you do cargo-install today (without your changes) 1.5.5 would be used anyway.
 
 Yes you are right but this isn't always guaranteed, there could be a chance (in future say) that you include a dependency (even if transitive) then with the current setup you might end up with a version that you don't intend to. This just pushes the baseline and ensures that doesn't happen

sagiegurari:
got it. good point.

Cargo.toml
```diff
@@ -59,7 +59,7 @@ home = "^0.5"
 indexmap = { version = "^1", features = ["serde-1"] }
 ignore = "^0.4"
 log = "^0.4"
-regex = "^1"
+regex = "1.5.5"
```
sagiegurari, line 62: i know its default in cargo, but its not the semver way. i'm not sure why cargo decided that by default ^ is implied even if not written, but i feels its not readable.
can we just add the ^ so it will be ^1.5.5

sagiegurari:
@Dylan-DPC can we just add the: ^ there? i know cargo does it for us, but its not the semver way and i feel its confusing for non rust people.

sagiegurari:
@Dylan-DPC I'll update the version there. in mean time, i'm merging this. thanks

Issue #374, opened 2020-02-09
For some executables, cntr-c will leave zombie processes


sagiegurari:
This is in to continue #355 since original issue was resolved, but additional flow reported by @mogenson was not and requires different handling.
There is a hacky workaround to use script executable for it in the meantime.

repro can be seen in https://github.com/mogenson/cargo-make-ctrl-c.git

MartinKavik:
Hi,
I think I have a similar problem with `Ctrl+C`. 
A `cargo-make` task is stopped by `Ctrl+C` but the process started by the associated command is shutting down and still writing to the terminal. The result is this:

![image](https://user-images.githubusercontent.com/18517402/162638442-c2c2c491-4cda-4240-aa1d-bb8d30ed9498.png)

The task is just a `cargo run` command; the executed app uses [tokio::signal::ctrl_c](https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html) to implement custom graceful shut down:
```toml
# ...
[tasks.mzoon]
description = "Run MZoon"
command = "cargo"
args = ["run", "--manifest-path", "../../crates/mzoon/Cargo.toml", "${@}"]
```

The first idea how to resolve it - add the `forward_signals` or `ignore_signals` property with a `bool` or `enum/array` value:
```toml
[tasks.mzoon]
ignore_signals = ["all"]   # <----
command = "cargo"
args = ["run", "--manifest-path", "../../crates/mzoon/Cargo.toml", "${@}"]
```
Then `cargo-make` would wait for the task to finish even after the user pressed `Ctrl+C`.

sagiegurari:
@MartinKavik thanks for the feedback.
The idea is interesting but i'm worried about use cases like the one that originally opened this issue which is a sub process that runs forever (such as a server) until you break it.

is there any signal you are getting that parent process died in your sub process? i guess the ctrlc you are not getting but i wonder whats there that prevents the process from being killed when the parent dies as well and how i can force kill those.

MartinKavik:
I'm not sure if it'll answer your question, but:

I've made a simple experiment - added `ctrlc = "3.2.1"` to `cargo-make`'s `Cargo.toml` and modified `run_command_get_output`:
```rust
/// Runs the requested command and return its output.
pub(crate) fn run_command_get_output(
    command_string: &str,
    args: &Option<Vec<String>>,
    capture_output: bool,
) -> io::Result<Output> {

    // --- NEW ---
[...]

PR #659, opened 2022-06-01
Double Ctrl+C

MartinKavik:
Related issue: #374 

Resolves the problem when the `cargo-make` binary exits too early (because of receiving Ctrl+C signal) but the task command is still running (either it's shutting down or becoming a zombie process).

The first Ctrl+C press writes `Shutting down...` to the terminal and waits for the command to shut down gracefully. The second press kills the command and waits for its exit status.

Tested manually with a new "playground repo" [makers_tester](https://github.com/MartinKavik/makers_tester) and with [MoonZoon](https://github.com/MoonZoon/MoonZoon) examples. `cargo-make` unit tests pass (run in Git hooks). Tested on Windows.

I'm not sure if the `loop` is the most efficient implementation but I didn't want to introduce some async runtimes or complex streaming/channels in the code.

P.S. `cargo-make` writes `[cargo-make] INFO - Build Done in 22.51 seconds.` when the task is finished. The `Build Done` text doesn't make too much sense when the task is a running server or something like that. Just fyi.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ cargo_metadata = "^0.14.0"
 ci_info = "^0.14.3"
 clap = "^3.1"
 colored = "^2"
+ctrlc = "^3.2.2"
 dirs-next = "^2"
 duckscript = "^0.7.1"
 duckscriptsdk = { version = "^0.8.10", default-features = false }
--- a/src/lib/command.rs
+++ b/src/lib/command.rs
@@ -169,8 +169,8 @@ pub(crate) fn run_command_get_output(
     };
 
     command.stdin(Stdio::inherit());
-    if !capture_output {
-        command.stdout(Stdio::inherit()).stderr(Stdio::inherit());
+    if capture_output {
+        command.stdout(Stdio::piped()).stderr(Stdio::piped());
     }
     info!("Execute Command: {:#?}", &command);
 
--- a/src/lib/command.rs
+++ b/src/lib/command.rs
@@ -9,7 +9,7 @@ mod command_test;
 
 use crate::logger;
 use crate::toolchain;
-use crate::types::{CommandSpec, Step};
+use crate::types::{CommandSpec, Step, UnstableFeature};
 use envmnt;
 use run_script;
 use run_script::{IoOptions, ScriptError, ScriptOptions};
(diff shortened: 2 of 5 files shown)


sagiegurari:
@MartinKavik first of all thanks a lot for the PR.
I got few concerns over the changes and its possibly just me, but I wonder maybe we should make this capability optional and turned on via config section in the makefile.toml.
By default, cargo-make will behave as today, but if turned on, we will get the new flow that you created.
This will allow us to avoid any blockers that might be introduced with such a change.
if your project requires it, you can set it in the makefile.toml and any contributor would have it turned on by default.
does it make sense? what do you think?

MartinKavik:
> maybe we should make this capability optional and turned on via config section in the makefile.toml

No problem for me, I already have some basic configs in most makefiles, e.g.:
```toml
[config]
default_to_workspace = false
min_version = "0.32.15"

[config.modify_core_tasks]
private = true
namespace = "default"

[tasks.xxx]
```
so I can add another field when needed.

---

> By default, cargo-make will behave as today, but if turned on, we will get the new flow that you created.
This will allow us to avoid any blockers that might be introduced with such a change.

Perhaps we can draw inspiration from Rust and introduce something like:
```toml
[config]
unstable_features = ["ctrl_c_handling"]
```
Just an idea, I don't have a strong opinion about it.

---

Should I try to implement this gate/configuration and somehow resolve codecov problems or do you want to do it by yourself?

sagiegurari:
> Should I try to implement this gate/configuration and somehow resolve codecov problems or do you want to do it by yourself?

your choice. I would love you to take it but i'm ok if you want me to push the PR as is and do the changes no top. just say what you prefer.

> unstable_features = ["ctrl_c_handling"]

interesting, on one hand, we can more things here easily, on the other hand, its less type safe.
I like it. lets go for it.

MartinKavik:
Added a new commit `unstable_features`.

Updated `makers_tester`'s makefiles for manual testing: https://github.com/MartinKavik/makers_tester/commit/2ee95483f6e296531f04505be6886063dfe99067

Implementation notes: 
- I tried to make _unstable features_ as typed as possible. 
- The function `UnstableFeature::to_env_name` could be optimized a bit to return `&'static str` instead of `String` but we would need to integrate a crate like `strum` to automatically derive env names from the enum variants and then maybe another crate to join env name parts during the compilation.
- I've chosen the implementation with env vars to add a bit more flexibility for users and to not need to painfully pass the config values through too many functions. 
- I've neither updated docs nor written new unit tests for the functionality - I think the tasks would be more suitable for you, if possible.

sagiegurari:
@MartinKavik amazing job. don't have any critical code comments, might do minor changes later but overall looks great.
I'll do so minor testing tomorrow and merge this.
the whole type safety you did for the unstable features is awesome. thanks a lot for that

sagiegurari:
@MartinKavik going to merge this in. looks awesome.
will probably do few minor nit changes and also fix the build by defining a new min rust version for it.
many thanks for the help here

Issue #341, opened 2019-12-23
Debian Packaging


mmstick:
Would you be interested in having debian packaging in the repository?

[Currently maintaining this in a fork for Pop!_OS](https://github.com/pop-os/cargo-make/tree/master/debian).

sagiegurari:
would love to have it.
so i can install it via apt-get? and how is it getting updated in case of a new cargo-make release?
(I got zero knowledge on how to publish to linux package manager repository)

mmstick:
You can on Pop!_OS, at least, since that's where it's now being packaged. It would allow anyone to build debian packages from the source repo, too. `ie: dpkg-buildpackage -b`

Any repository in the Pop!_OS GitHub repository that contains a debian directory is automatically packaged by our CI.

sagiegurari:
cool. maybe you can document it in the installation section?

John15321:
@mmstick 
@sagiegurari 
Could we use `cargo-deb` for that?
https://crates.io/crates/cargo-deb
"This is a [Cargo](https://doc.rust-lang.org/cargo/) helper command which automatically creates binary [Debian packages](https://www.debian.org/doc/debian-policy/ch-binary.html) (.deb) from Cargo projects."

sagiegurari:
ya probably but I'm not familiar with the entire publishing process to debian. need to learn a bit on it.

but a good start would be to create a deb file and have it part of the binary release artifacts

John15321:
> ya probably but I'm not familiar with the entire publishing process to debian. need to learn a bit on it.
> 
> but a good start would be to create a deb file and have it part of the binary release artifacts

I would really like to contribute to your package because I really like it so If you would like I can take of it 😄

John15321:
> that would be great. if you could create a task in the makefile to create the deb and have it run in the release build and upload it would be awesome

Sure 😄 I will start tomorrow and ask you if I don't know something

John15321:
@sagiegurari 

Hey Im trying to build it locally (on an ubuntu vm, its on arm)

And Im getting this compilation error (I set my rustc version to 1.56.0):

```bash
[cargo-make] INFO - cargo make 0.35.13
[cargo-make] INFO - Project: cargo-make
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: ci-flow
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: legacy-migration
[...]

PR #665, opened 2022-06-25
341 add debian packaging to the release files

John15321:
Resolves https://github.com/sagiegurari/cargo-make/issues/341

Adds a `.deb. package to the release

### DIFF
--- a/.github/workflows/deb.yml
+++ b/.github/workflows/deb.yml
@@ -1,29 +1,32 @@
-
-
-name: deb
-on: [push, pull_request]
+name: Release deb package
+on:
+  push:
+    tags:
+      - '*'
 env:
   CLICOLOR_FORCE: 1
 jobs:
-  deb:
-    name: Build deb package
+  release-deb:
+    name: Build and release a .deb package
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
     steps:
-    - name: Install musl tools
-      run: sudo apt-get install musl-tools -y
     - name: Checkout
       uses: actions/checkout@v2
+    - name: Install rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        profile: minimal
+        override: true
     - name: Install cargo-deb
       run: cargo install cargo-deb
-    - name: Create a deb package
+    - name: Build .deb package
       run: cargo deb
-    - name: Upload Binaries
+    - name: Upload .deb package
       uses: svenstaro/upload-release-action@v1-release
       with:
         repo_token: ${{ secrets.GITHUB_TOKEN }}
-        file: target/x86_64-unknown-linux-musl/debian/cargo-make_0.35.13_amd64.deb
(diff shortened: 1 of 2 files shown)


sagiegurari:
thanks a lot, will check it.

sagiegurari:
@John15321 this looks cool, but can we update the deb file to include the platform on which it is supported?
i'm guessing i can't take that deb file and use it on arm, right?

John15321:
> @John15321 this looks cool, but can we update the deb file to include the platform on which it is supported?
> i'm guessing i can't take that deb file and use it on arm, right?

The generated Deb file automatically has the version number of the package and it's architecture in the filename. At least it should

John15321:
> @John15321 this looks cool, but can we update the deb file to include the platform on which it is supported?
> i'm guessing i can't take that deb file and use it on arm, right?



As you can see here it says it's `amd64`

sagiegurari:
ah sorry for missing that. this is great. i'll take a closer look this week. sorry for the delay

John15321:
> ah sorry for missing that. this is great. i'll take a closer look this week. sorry for the delay

No problem ☺️ I'm happy to help

sagiegurari:
@John15321 i've been thinking about this one a lot.
1. why put these commands in the yaml and not in the makefile.toml? this way i can test it without github actions
2. if we are putting it in the toml, why not in the internal toml so others can use it as well. maybe in the rust.toml
3. if we are pushing it in, lets make it a flow with pre/post empty tasks so people can customize the flow
4. you said you thought about later doing it for fedora and such... so maybe we can even push it even further to have conditions based on dist and call the right task. such conditions are not supported by cargo-make (i can check env, os and such but not which linux you are running on) so you will have to build a condition using duckscript (or routing...)

so, something like this (not tested, just concept)

```toml
[tasks.package-native-flow]
dependencies = ["pre-package-native", "package-native", "post-package-native"]

[tasks.pre-package-native]
alias = "empty"

[tasks.post-package-native]
alias = "empty"

[tasks.package-native]
condition = { platforms = ["linux"] }
script = '''
#!@duckscript

# check which dist and call right task, for example lets say we found its a debian and we created a var named debian)
if ${debian}
  cm_run_task cargo-deb
end
'''

# since this is a cargo command, cargo-make will install cargo-deb if needed automatically.
[tasks.cargo-deb]
command = "cargo"
args = ["deb"]
```

the above is not tested and missing few pieces. 
wondering what you think?

sagiegurari:
@John15321 this PR got closed since i deleted the branch.
can you rebuild the PR again please but also incorporate the suggestions in my previous message?

John15321:
> @John15321 this PR got closed since i deleted the branch. can you rebuild the PR again please but also incorporate the suggestions in my previous message?

Im sorry I forget and didn't have time. I will do it this weekend

sagiegurari:
no rush. i just don't want this to be lost by mistake

PR #672, opened 2022-07-21
Allow building with native-tls instead of rustls

jirutka:
Bundling (static linking) TLS library is a very bad idea from a security perspective and linux distributions try to avoid it.

See-Also: https://github.com/sagiegurari/duckscript/pull/258


jirutka:
FYI: It’s really not a good idea to use the same naming schema for branches and tags (`0.35.14`).

PR #688, opened 2022-08-06
env variable lift order requirement.

### LINKED ISSUE #685: Env variable interpolation stops working when extending
### Describe The Bug

When redefining environmental variables, from a Makefile that we extend from and execute any task `@@split()` no longer works.

### To Reproduce

Minimal Reproducible Example:

extend.toml

```toml
extend = { path = "base.toml" }

[env]
CARGO_TEST_HACK_FLAGS_BASE = "--workspace --feature-powerset --exclude-features default --optional-deps eyre,anyhow"
CARGO_TEST_HACK_FLAGS = "${CARGO_TEST_HACK_FLAGS_BASE} --group-features anyhow,eyre --group-features std,hooks --group-features spantrace,futures,small"


[tasks.test2]
command = "echo"
args = ["@@split(CARGO_TEST_HACK_FLAGS, )"]
```

base.toml

```toml
[config]
default_to_workspace = false

[config.modify_core_tasks]
private = true
namespace = "default"

[env]
CARGO_TEST_HACK_FLAGS = "--workspace --feature-powerset"


[tasks.test1]
command = "echo"
args = ["@@split(CARGO_TEST_HACK_FLAGS, )"]
```

```console
> cargo make --makefile extend.toml test2
[cargo-make] INFO - cargo make 0.35.15
[cargo-make] INFO - Build File: extend.toml
[cargo-make] INFO - Task: test2
[cargo-make] INFO - Running Task: default::legacy-migration
[cargo-make] INFO - Execute Command: "echo" "--workspace --feature-powerset --exclude-features default --optional-deps eyre,anyhow" "--group-features" "anyhow,eyre" "--group-features" "std,hooks" "--group-features" "spantrace,futures,small"
--workspace --feature-powerset --exclude-features default --optional-deps eyre,anyhow --group-features anyhow,eyre --group-features std,hooks --group-features spantrace,futures,small
[cargo-make] INFO - Build Done in 0.10 seconds.
> cargo make --makefile extend.toml test1
[cargo-make] INFO - cargo make 0.35.15
[cargo-make] INFO - Build File: extend.toml
[cargo-make] INFO - Task: test1
[cargo-make] INFO - Running Task: default::legacy-migration
[cargo-make] INFO - Execute Command: "echo" "--workspace --feature-powerset --exclude-features default --optional-deps eyre,anyhow" "--group-features" "anyhow,eyre" "--group-features" "std,hooks" "--group-features" "spantrace,futures,small"
--workspace --feature-powerset --exclude-features default --optional-deps eyre,anyhow --group-features anyhow,eyre --group-features std,hooks --group-features spantrace,futures,small
[cargo-make] INFO - Build Done in 0.11 seconds.
```

indietyp:
This Pull Request adds the ability (when loading) to reorder the environment variables based on usage, removing the requirement that variables need to follow a specific definition order.

Even though this only does the reordering on `merge_env`. 

This change will always be applied because `merge_env` is called in `merge_base_config_and_external_config`.

Closes #685


indietyp:
I'd like to create some acceptance tests, how would I go about doing that?

src/lib/descriptor/mod.rs
```diff
@@ -14,69 +14,175 @@ mod cargo_alias;
...
-use crate::io;
-use crate::plugin::descriptor::merge_plugins_config;
-use crate::scriptengine;
-use crate::types::{
-    Config, ConfigSection, EnvFile, EnvFileInfo, EnvValue, Extend, ExternalConfig, ModifyConfig,
-    Task,
-};
-use crate::version;
-use envmnt;
+use std::collections::HashSet;
```
sagiegurari, line 17: what is this custom 'use' order? i put everything and let rustfmt (cargo make format) order it automatically.

src/lib/descriptor/mod.rs
```diff
@@ -14,69 +14,175 @@ mod cargo_alias;
...
+use crate::plugin::descriptor::merge_plugins_config;
+use crate::types::{
+    Config, ConfigSection, EnvFile, EnvFileInfo, EnvValue, EnvValueConditioned, EnvValueDecode,
+    EnvValuePathGlob, Extend, ExternalConfig, ModifyConfig, Task,
+};
+use crate::{io, scriptengine, version};
+
+static RE_VARIABLE: Lazy<Regex> = Lazy::new(|| Regex::new(r"\$\{.*}").unwrap());
+
+fn merge_env_depends_on_extract(val: &str) -> Vec<&str> {
```
sagiegurari, line 40: can we move the entire env loading/ordering logic to a new file? this file is big as it is

src/lib/descriptor/mod_test.rs
```diff
@@ -9,18 +8,60 @@ fn merge_env_both_empty() {
...
+    map1.insert("env2".to_owned(), EnvValue::Value("${env1}".to_owned()));
+
+    let mut map2 = IndexMap::new();
+    map2.insert("env1".to_owned(), EnvValue::Value("x".to_owned()));
+
+    let output = merge_env(&map1, &map2).expect("should have no cycle");
+    assert_eq!(output.len(), 3);
+    assert_eq!(output.get_index(0).unwrap().0.as_str(), "env1");
+    assert_eq!(output.get_index(1).unwrap().0.as_str(), "env2");
+    assert_eq!(output.get_index(2).unwrap().0.as_str(), "env3");
```
sagiegurari, line 42: i get why you want that, but if the env defined:
env3
env2
it should keep that order. it is up to the user while writing a makefile to put the env vars in order.

src/lib/descriptor/mod.rs
```diff
@@ -14,69 +14,175 @@ mod cargo_alias;
...
+    let mut graph: GraphMap<&str, (), _> = DiGraphMap::new();
+
+    let keys: HashSet<_> = combined.iter().map(|(key, _)| *key).collect();
+    for key in keys {
+        graph.add_node(key);
     }
 
-    for (key, value) in extended.iter() {
-        let key_str = key.to_string();
+    debug!("initial graph: {graph:?}");
```
sagiegurari, line 112: lets just make sure the min version of rust that cargo-make supports, supports this feature (i think it does...)

src/lib/descriptor/mod.rs
```diff
@@ -14,69 +14,175 @@ mod cargo_alias;
...
+                }
+            }
 
-                            merged.insert(key_str, EnvValue::Profile(merged_sub_env));
-                        }
-                        _ => {
-                            merged.insert(key_str, value_clone);
-                        }
-                    };
+            return Err(err);
```
sagiegurari, line 149: breaking the build due to a cycle is a new behavior. I hope this won't break other projects.

sagiegurari (changes requested):
wow! amazing job and super fast. love it.
gave few comments. i feel this might be a bit risky so lets see if we can try to minimize surprises to existing users.

sagiegurari:
@indietyp amazing work. i put some comments.
don't worry about the #ignore since cargo-make tests run in 2 phases
1. all non ignored tests in parallel
2. all ignored test as single thread (since they may impact each other)
so ALL tests will run if you just do 'cargo make'

src/lib/descriptor/mod.rs
indietyp: Oh yes, do you use a custom rustfmt config by any chance? There was no `rustfmt.toml` in the project directory, so it must've picked up my default configuration.

src/lib/descriptor/mod.rs
indietyp: I don't think so. I believe, in general, this is an improvement because it tells the user specifically why a configuration won't work. Pre PR, this would just silently error out, meaning it would lead to more frustration if that makes sense.

src/lib/descriptor/mod_test.rs
indietyp: Let us assist the user in being able to reorder the environment because this is what we already kind of do (on `merge_env`, we replace existing values). This is because we're handling configuration, not code. After all, I think in all cases, if I define a variable - independent of the order. I'd want them to be appropriately handled. This is how tools like `terraform` (https://www.terraform.io/language/values/outputs) work.

This also simplifies implementation, as we do a single passthrough and topologically sort.

This change is backward compatible, so old workflows won't break (evident by the CI not breaking) but instead enhances the current capabilities by allowing a more declarative approach.

indietyp: I think that, in general, if someone writes:

```
ENV2 = "${ENV3}"
ENV3 = "2"
```

They did not mean a global variable `ENV3`, then reassigned the value. I think this most likely leads to more confusion than it helps. The general mindset is that if I define an env variable in my block, then my program will use that variable. Otherwise, it might lead to unexpected results - and frustration - instead.

indietyp: I'd happily add a comment about that behavior in the `README.md`, but I think it's a valuable addition.

src/lib/descriptor/mod.rs
sagiegurari: I don't use any custom formatting. only default one on purpose.

src/lib/descriptor/mod_test.rs
sagiegurari: maybe writing it in the README is a good idea. I have a long section with subsections on the env loading since its so complex and got so many capabilities already (env file, env on cli, os env, env block, profiles, env on task level, env scripts, .... )

src/lib/descriptor/mod.rs
sagiegurari: agree with approach. just concerned. later, i'm getting lots of issues from unhappy people :)

indietyp: I see. This is why specifically search for strongly connected components so that people can see **where** a cycle is.

So if you declare

```
ENV2 = ${ENV1}
ENV1 = ${ENV2}
```

then it will tell you: cycle detected: ENV2 -> ENV1 -> ENV2.

I think this should help minimize the number of issues. Maybe I can clarify the issue message even more, to redirect to the README for a link explaining how this works.

src/lib/descriptor/mod.rs
indietyp: would it be okay to just create an empty `rustfmt.toml` in the root directory? That way - even if people (like me) - have a custom configuration. It won't take precedence over the default formatting rules.

src/lib/descriptor/mod.rs
sagiegurari: cool

indietyp:
I have reworked the section about env variables in the `README.md`. I just need to add the error-code docs and improve the coverage, and I think that's about it!

sagiegurari:
@indietyp i'll review it again once you resolve the conflicts.
i've released a new version and seems that they conflict with your changes.

sagiegurari:
@indietyp insane work. it will take me time to review it properly.
i did see you wrote a lot in the readme including stuff i was too lazy to write, so thanks a lot. but, the readme is generated from the content.md and nav.md so i'll lose all your great docs :) so you need to move it to there.

I need to understand everything you did there, since i saw logic to understand depends on per type and even for scripts added a new field. i'm for it, but it will just take me time to fully understand all changes there to ensure its ok.

indietyp:
Yea, I totally get it, thanks for looking into it, I didn't know that the `README.md` was generated from the `content.md`, so I will move my new content there. :D

README.md
```diff
@@ -1343,18 +1344,145 @@ We run task **3** the output would be:
...
+
+`cargo-make` supports the use of glob syntax to find all files and directories in a given directory. The list of files will be joined using `;` during execution.
+
+```toml
+PATH_GLOB = { glob = "./src/**/mod.rs", include_files = true, include_dirs = false, ignore_type = "git" }
+```
+
+##### Conditional
+
+`cargo-make` supports conditional variables, which are set to the `value` specified if the `condition` evaluates to true. To learn more about conditions, refer to [this chapter]
```
sagiegurari, line 1415: lets fix [this chapter] to the right link :)

README.md
```diff
@@ -1343,18 +1344,145 @@ We run task **3** the output would be:
...
+
+##### Unset
+
+Variables can be unset, which will unset their value and reset it to an empty string.
```
sagiegurari, line 1419: Let's simplify here since its not setting it to empty string. its removing it.
so maybe just write: "Variables can be unset." as i think its enough

README.md
```diff
@@ -1343,18 +1344,145 @@ We run task **3** the output would be:
...
+
+```toml
+VARIABLE = {unset = true}
+```
+
+#### Note about Ordering
+
+Environmental variables in `cargo-make`, create implicit dependencies between each other and are handled more in a declarative approach. Therefore ordering of variables is not necessarily the same between definition and evaluation.
```
sagiegurari, line 1427: this entire section is too long and detailed. lets just explain that order is important but also dependencies may impact it.
i don't think we need to expand too much what happens if there is only basic implementation if its not really like that now.
this includes the native/simple sub sections.

however, i would add that this logic is only for env block and not for task env block.

README.md
```diff
@@ -1343,18 +1344,145 @@ We run task **3** the output would be:
...
+
+##### Simple
+
+The most ordinary one is the definition of a simple `KEY=Value` pair, which is reminiscent of tools like [dotenv](https://www.npmjs.com/package/dotenv) and [bash scripts](https://www.gnu.org/software/bash/). Values can use other variables as values, which are interpolated at runtime, using the `${variable}` syntax.
+
+> **Note:** The `$variable` syntax is currently not supported!
+
+> **Note:** Extended interpolation like `${variable:-default}` are currently unsupported!
+
+```toml
```
sagiegurari, line 1371: string example?

README.md
```diff
@@ -1383,42 +1511,10 @@ DEV = true
...
-
-All environment variables defined in the env block and in the [default Makefile.toml](https://github.com/sagiegurari/cargo-make/blob/master/src/lib/descriptor/makefiles/stable.toml) will be set before running the tasks.<br>
-To unset an environment variable, use the **MY_VAR = { unset = true }** syntax.<br>
-See more on profile based environment setup in the [profile environment section](#usage-profiles-env)
-
 <a name="usage-env-task"></a>
 #### Task
-Environment variables can be defined inside tasks using the env attribute, so when a task is invoked (after its dependencies), the environment variables will be set, for example:
+
+Environmental variables can be set in the scope of the task and will be merged (and reordered) with the global environment when that task gets executed. This means that the evaluation of environmental variables takes place after all dependencies have run but before the task itself will run.
```
sagiegurari, line 1517: we support reordering on task level as well?

README.md
```diff
@@ -1432,8 +1528,6 @@ echo var: ${SOME_ENV_VAR}
 '''
 ```
 
-In task level, environment variables capabilities are the same as in the [global level](#usage-env-config).
```
sagiegurari, line 1435: why remove this?

README.md
```diff
@@ -1464,7 +1555,9 @@ ENV2_TEST=TEST2
 ENV3_TEST=VALUE OF ENV2 IS: ${ENV2_TEST}
 ```
 
-Env files can also be defined globally in the Makefile.toml via **env_files** attribute as follows:
+Paths to environment files can also be defined globally in the `env_files` key of the `Makefile.toml`, which will be loaded in the order they are defined. All paths are relative to the `Makefile.toml` directory containing them.
+
+> **Note:** `env_files` can only be used on a task level, be aware that relative paths will instead be relative **to the current working directory**
```
sagiegurari, line 1560: what is this note? its supported on global and task. why did you write only on task?

README.md
```diff
@@ -1492,22 +1586,13 @@ env_files = [
 ]
 ```
 
-In this example, **profile.env** is only loaded in case the runtime profile is **development**.<br>
-*More on profiles in the [profiles](#usage-profiles) section.*<br>
-<br>
-Relative paths are relative compared to the toml file that declared them and not to the current working directory.<br>
-<br>
-The same **env_files** attribute can be defined on the task level, however relative paths on the task level are relative to the current working directory.<br>
```
sagiegurari, line 1500: why is this block removed?

README.md
```diff
@@ -3831,6 +3920,31 @@ Tasks which only install some dependency but do not invoke any command start wit
 install_crate = { rustup_component_name = "rust-src" }
 ```
 
+<a name="error-codes"></a>
+## Error Codes
+
+<a name="e001"></a>
+### E001: Cycle Detected
```
sagiegurari, line 3927: Maybe 'Environment Variables Cycle Detected' since other cycles are possible (task dependencies for example)

sagiegurari (changes requested):
@indietyp it looks great.
I wrote few comments, mostly about the docs actually.

README.md
```diff
@@ -112,6 +112,7 @@
     * [Global Configuration](#cargo-make-global-config)
```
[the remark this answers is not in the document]
indietyp: Oh yes, I was pretty busy this week, so I haven't been able to move the documentation.

README.md
```diff
@@ -1343,18 +1344,145 @@ We run task **3** the output would be:
...
 * [Command Line](#usage-env-cli)
 * [Env File](#usage-env-file)
 * [Global](#usage-env-global)
 
+<a name="env-declaration"></a>
+#### Declaration
+
+There are multiple ways to declare environmental variables, all of which are suited for specific suitcases.
+
+##### Simple
```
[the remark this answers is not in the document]
indietyp: Markdown supports up to 6 levels, rendering depends highly on the renderer (here GitHub).

image

README.md
indietyp: `only` should've been `also`, thanks for the catch

README.md
indietyp: My bad, this only works on env blocks. I thought that `merge_env` was getting called in `environment::set_env_for_config`, but I was mistaken.

My question would be: do we want re-ordering at the last stage? So when we run an actual task (this would mean we would need to call `merge_env` in `run_task_with_options`, if I understood the code correctly.

sagiegurari:
@indietyp there are some CI errors to look at. I really want to merge this into the new cargo-make version

indietyp:
@sagiegurari same here, but I wanted to have some clarification: https://github.com/sagiegurari/cargo-make/pull/688#discussion_r944811401 and have been waiting for an answer.

README.md
indietyp: It wasn't. Just reworded into:

Paths to environment files can also be defined globally in the `env_files` key of the `Makefile.toml`, which will be loaded in the order they are defined. All relative paths are relative to the directory containing the `Makefile.toml` they were defined in.

> **Note:** `env_files` can also be used on a task level, be aware that relative paths will instead be relative **to the current working directory**

README.md
sagiegurari: na, i think we can document that we are only doing it for the env makefile level and not task level if it simplifies things.

sagiegurari:
sorry missed it.

README.md
indietyp: @sagiegurari I have a specific question about this. Let's say I have 3 tasks:

Makefile:
sets `env1` to `1`

Task A:
sets `env1` to `2`

Task B:
sets `env1` to `3`

Task C:
get `env1`
depends_on: [Task A, Task B]

Would `env1` for Task C be `3` in that case? It looked like env variables are not rolled back once a previous task sets them.

sagiegurari: yes, it would be 3. they are never rolled back. 
the whole point is that tasks allow you to modify the env and affect the next tasks that come.
for example one task may setup the env based on a condition block.

indietyp: I gotcha that makes sense.

docs/_includes/content.md
```diff
@@ -1213,20 +1213,151 @@ We run task **3** the output would be:
...
+LIST_VALUE = [ "VALUE1", "VALUE2", "VALUE3" ]
+```
+
+##### Script
+
+`cargo-make` supports the use of simple scripts. The output of the said script will then determine the value of the environmental variable.
+
+The script's object has two additional arguments: `multiline` and `depends_on`. If `multiple` is set to `true`, the supplied script will be evaluated as a script with multiple lines. `depends_on` is a list of environmental variables this script depends on, which is taken into account during reordering if unset `cargo-make` will try to guess the variables used during reordering.
+
+> **Note:** This uses `sh` for script evaluation ([spec](https://pubs.opengroup.org/onlinepubs/9699919799/)), other runners like `duckscript`, `rust`, etc. are **not** supported.
```
sagiegurari, line 1265: instead of:
This uses `sh` for script evaluation ([spec](https://pubs.opengroup.org/onlinepubs/9699919799/))

please write:
This uses the default OS command runner (cmd for windows, sh for the rest)

docs/_includes/content.md
```diff
@@ -1255,42 +1386,16 @@ DEV = true
...
-```toml
-PATH_GLOB = { glob = "./src/**/mod.rs", include_files = true, include_dirs = false, ignore_type = "git" }
-```
-* Key and a structure holding the value (can be an expression) and optional condition which must be valid in order for the environment variable to be set
+Environmental variables can be set in the scope of the task and will be merged with the global environment when that task gets executed. This means that the evaluation of environmental variables takes place after all dependencies have run but before the task itself will run.
 
-All environment variables defined in the env block and in the [default Makefile.toml](https://github.com/sagiegurari/cargo-make/blob/master/src/lib/descriptor/makefiles/stable.toml) will be set before running the tasks.<br>
-To unset an environment variable, use the **MY_VAR = { unset = true }** syntax.<br>
-See more on profile based environment setup in the [profile environment section](#usage-profiles-env)
+> **Note:** Reordering of task variables with global variables will **not** take place, tasks will simply overwrite previously declared variables.
```
sagiegurari, line 1394: from:
take place, tasks will

to:
take place.
Tasks will

docs/_includes/content.md
```diff
@@ -1336,7 +1436,9 @@ ENV2_TEST=TEST2
 ENV3_TEST=VALUE OF ENV2 IS: ${ENV2_TEST}
 ```
 
-Env files can also be defined globally in the Makefile.toml via **env_files** attribute as follows:
+Paths to environment files can also be defined globally in the `env_files` key of the `Makefile.toml`, which will be loaded in the order they are defined. All relative paths are relative to the directory containing the `Makefile.toml` they were defined in.
+
+> **Note:** `env_files` can also be used on a task level, be aware that relative paths will instead be relative **to the current working directory**
```
sagiegurari, line 1441: from:
task level, be aware

to:
task level.
Be aware

sagiegurari:
@indietyp amazing work. only few small comments on the docs. so once done i'll merge it in.
thanks a lot

sagiegurari:
@indietyp sorry for the delay, i was not available. i'll recheck it.

docs/_includes/content.md
```diff
@@ -1213,20 +1213,151 @@ We run task **3** the output would be:
...
 <a name="usage-env"></a>
 ### Environment Variables
-cargo-make enables you to defined environment variables in several ways.<br>
-Environment variables can later be used in commands, scripts, conditions, functions and more, so it is important to have a powerful way to define them for your build.
+`cargo-make` enabled the definition of environmental variables in several ways, which can later be accessed throughout task execution.
+
+Because environmental variables play a significant role in `cargo-make`, it provides multiple declarative ways to provide them at different levels of granularity.
 
+* [Declaration](#env-declaration)
+* [Note about Ordering](#env-note-about-ordering)
```
sagiegurari, line 1224: feels like this should come as last sub section and not second.

docs/_includes/content.md
```diff
@@ -1336,7 +1436,9 @@ ENV2_TEST=TEST2
 ENV3_TEST=VALUE OF ENV2 IS: ${ENV2_TEST}
 ```
 
-Env files can also be defined globally in the Makefile.toml via **env_files** attribute as follows:
+Paths to environment files can also be defined globally in the `env_files` key of the `Makefile.toml`, which will be loaded in the order they are defined. All relative paths are relative to the directory containing the `Makefile.toml` they were defined in.
+
+> **Note:** `env_files` can also be used on a task level. Be aware that relative paths will instead be relative **to the current working directory**
```
sagiegurari, line 1441: instead of
```**to the current working directory**```
change to
```to the **current working directory**```

File: docs/_includes/nav.md [hunk 22]
```diff
@@ -24,6 +24,8 @@
         * [Platform Override](#usage-platform-override)
         * [Extend Attribute](#usage-task-extend-attribute)
     * [Environment Variables](#usage-env)
+        * [Declaration](#env-declaration)
+        * [Note about Ordering](#env-note-about-ordering)
```
sagiegurari, line 28: move as last sub section instead of second

sagiegurari (changes requested):
all the comments i put are minor and i can fix myself if you want.
but the biggest issue that i suddenly found (based on the docs you wrote) is that the support for ${env:default} is dropped?
we can't drop that as it will break peoples projects.

docs/_includes/content.md
```diff
@@ -1213,20 +1213,151 @@ We run task **3** the output would be:
...
+
+There are multiple ways to declare environmental variables, all of which are suited for specific suitcases.
+
+##### Simple
+
+The most ordinary one is the definition of a simple `KEY=Value` pair, which is reminiscent of tools like [dotenv](https://www.npmjs.com/package/dotenv) and [bash scripts](https://www.gnu.org/software/bash/). Values can use other variables as values, which are interpolated at runtime, using the `${variable}` syntax.
+
+> **Note:** The `$variable` syntax is currently not supported!
+
+> **Note:** Extended interpolation like `${variable:-default}` is currently unsupported!
```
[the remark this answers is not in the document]
indietyp: Could it be that envmnt supports non-standard interpolation through: `:default` instead?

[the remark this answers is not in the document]
sagiegurari: @indietyp you are right! it has been so long. envmnt actually supports it so i'll now add it to cargo-make.
so this can be ignored and i'll drop the comment

sagiegurari:
@indietyp thanks a lot for the PR. I know it has been a long ride :) so sorry about that.
i'll do the minor fixes i wrote in my comments and i've added the ${value:default} capability (its actually already supported in envmnt, just needed to turn it on).

docs/_includes/content.md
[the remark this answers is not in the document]
indietyp: I am thinking of creating a new PR to add support for the full range of interpolation as per POSIX spec: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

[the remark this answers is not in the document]
sagiegurari: to envmnt? sounds good. let me check that link and we will limit it only to ${var} and not $var right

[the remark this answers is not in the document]
indietyp: Yep correct!

PR #697, opened 2022-08-16
Extent format-toml with optional arguments

phoracek:
With this patch, it is possible to extend format-toml target with arbitrary space-separated arguments.

One of the use-cases for this may be running a read-only toml check, which could fail CI in case a TOML file is not formatted correctly:

```toml
[tasks.check-toml]
env = { CARGO_MAKE_FORMAT_TOML_ARGS = "--check" }
run_task = { name = "format-toml" }
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1121,18 +1121,18 @@ dependencies = [
 
 [[package]]
 name = "serde"
-version = "1.0.142"
+version = "1.0.143"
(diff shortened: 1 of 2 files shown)


sagiegurari:
@phoracek thanks a lot. once the CI finishes, i'll merge it into a new dev branch.

sagiegurari:
@phoracek merged. thanks a lot

phoracek:
Thank you o/

PR #859, opened 2023-06-03
fix: --hide-uninteresting doesn't work

xxchan:


sagiegurari:
@xxchan thanks a lot. will release it soon

PR #866, opened 2023-06-08
feat: group aliases in --list-all-steps

### LINKED ISSUE #862: merge together alias tasks in `--list-all-steps`

Change
```
Misc
----------
l - Description
logs - Description
```
To
```
Misc
----------
logs - Description [aliases: l]
```

Would that be better? 

### Code Sample

```rust
/// paste code here
```

xxchan:
close #862

### DIFF
--- a/src/lib/cli_commands/list_steps.rs
+++ b/src/lib/cli_commands/list_steps.rs
@@ -20,8 +20,8 @@ pub(crate) fn run(
     output_file: &Option<String>,
     category: Option<String>,
     hide_uninteresting: bool,
-) -> u32 {
-    let (output, count) = create_list(&config, output_format, category, hide_uninteresting);
+) {
+    let output = create_list(&config, output_format, category, hide_uninteresting);
 
     match output_file {
         Some(file) => {
@@ -30,8 +30,6 @@ pub(crate) fn run(
         }
         None => print!("{}", output),
     }
-
-    count
 }
 
 /// Panics if task does not exist.
@@ -40,9 +38,7 @@ pub(crate) fn create_list(
     output_format: &str,
     category_filter: Option<String>,
     hide_uninteresting: bool,
-) -> (String, u32) {
-    let mut count = 0;
-
+) -> String {
     // category -> actual_task -> description
     let mut categories: BTreeMap<String, BTreeMap<String, String>> = BTreeMap::new();
     // actual_task -> aliases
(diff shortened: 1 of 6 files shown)


src/lib/cli_commands/list_steps.rs
```diff
@@ -33,29 +34,29 @@ pub(crate) fn run(
...
-    let markdown = single_page_markdown
-        || output_format == "markdown"
-        || output_format == "markdown-sub-section";
-    let just_task_name = output_format == "autocomplete";
-
-    let mut categories = BTreeMap::new();
 
-    if single_page_markdown {
-        buffer.push_str(&format!("# Task List\n\n"));
-    }
```
xxchan, line 55: Moved this section below to make the 2-phase more clear.

src/lib/cli_commands/list_steps.rs
```diff
@@ -89,6 +90,14 @@ pub(crate) fn create_list(
 
             count = count + 1;
 
+            if &actual_task_name != key {
+                aliases
+                    .entry(actual_task_name)
+                    .or_default()
+                    .insert(key.clone());
+                continue;
+            }
```
xxchan, line 99: I put it here so that `count` doesn't change. But actually I don't know why `count` is needed. It seems it's only used for testing. 

I think a better approach for testing is simply assert the output string against a snapshot. And there's a good tool [expect_test](https://docs.rs/expect-test/latest/expect_test/) for such purpose, which supports auto-update the results. If you are interested, I can refactor the tests for you to take a look.

xxchan: Now I only manually tested with `cargo run --bin makers -- --list-all-steps`

src/lib/cli_commands/list_steps.rs
```diff
@@ -117,17 +126,25 @@ pub(crate) fn create_list(
...
-                Some(value) => tasks_map.append(value),
-                _ => (),
-            };
-
-            tasks_map.insert(key.clone(), text.clone());
-            categories.insert(category, tasks_map);
+            categories
+                .entry(category)
+                .or_default()
+                .insert(key.clone(), text);
```
xxchan, line 132: style refactor

src/lib/execution_plan.rs
xxchan, line 1: Added some comments

xxchan:
I've refactored the tests. The first commit refactored them with placeholder and the output is like:

image


The second commit updates it with `env UPDATE_EXPECT=1 cargo test`

sagiegurari:
@xxchan awesome work! thanks a lot. merging to a dev branch first.

PR #869, opened 2023-06-10
add "default" for --hide-uninteresting

xxchan:


sagiegurari:
actually i think the default task is the most interesting task.
its the one triggered when you don't state a task name.

xxchan:
But we don't need to invoke it with the task name `cargo make default`? 😄
I thought hiding e.g., "init"/"end" is due to the same reason. (Aren't they
also "interesting" in this sense?)

When I `--list-all-steps` I want to see the tasks I may want to invoke.
Anyway, you decide.

Sagie Gur-Ari ***@***.***> 于2023年6月11日周日 05:36写道：

sagiegurari:
init/end are hooks. so decided to hide all hooks. things that happen before/after something.
so you can focus on the tasks that do actual work and not pre->preparation or post->cleanups.
default does something and probably the most common one, so if i list it, i see what default does based on its description.
so i'll know what to expect when running 'cargo make' without args.

PR #880, opened 2023-06-26
Arch Linux installation

CosminGGeorgescu:
Thought that it would be helpful to have this here


sagiegurari:
@CosminGGeorgescu yes it would, but can you update the content.md and generate the readme from it?
since readme is an output of that so any change to it directly would be deleted

PR #897, opened 2023-07-18
crateinfo: do not sort workspace members

### LINKED ISSUE #833: Workspace children do not get executed in order defined in `Cargo.toml` members
### Describe The Bug
The expected behaviour is that the order of execution for a workspace follows the order of the `members` array in `Cargo.toml`. In version v0.36.6 this is broken and appears to follow alphabetical order instead, v0.35.x works without issue

### To Reproduce

1. Have a workspace project with multiple members using cargo-make
2. Execute flow with v0.35 and observe execution order following `members`
3. Execute flow with v0.36 and observe execution order not following `members`

### Code Sample

I was running into this issue on this repo: https://github.com/WootingKb/wooting-analog-sdk I just made the change in the CI to use v0.35 to fix my testing flow. Running `cargo make test-flow` with v0.35 runs and completes successfully, but with `v0.36.6` it fails due to it executing tests on the `wooting-analog-sdk` member before running `wooting-analog-test-plugin` which is a dependency for the `wooting-analog-sdk` tests.

pskrgag:
Official doc states following

> The order of the members is defined by the member attribute in the workspace Cargo.toml.

, but it was broken in commit fe5ca3aba1050d0bce39663e0bd270d2a7d83049.

Remove sorting and add new test-case to catch such problems in future

Closes  #833

Signed-off-by: Pavel Skripkin <paskripkin@gmail.com>

### DIFF
--- a/src/lib/environment/crateinfo.rs
+++ b/src/lib/environment/crateinfo.rs
@@ -197,7 +197,6 @@ fn add_members(crate_info: &mut CrateInfo, new_members: Vec<String>) {
 fn dedup_members(crate_info: &mut CrateInfo) {
     if let Some(ref mut workspace) = crate_info.workspace {
         if let Some(ref mut members) = workspace.members {
-            members.sort();
             members.dedup();
         }
     }
--- a/src/lib/environment/crateinfo_test.rs
+++ b/src/lib/environment/crateinfo_test.rs
@@ -50,6 +50,24 @@ fn add_members_workspace_new_members_with_data() {
 }
 
 #[test]
+fn add_members_workspace_non_alpha_ordered() {
+    let mut crate_info = CrateInfo::new();
+
+    crate_info.workspace = Some(Workspace::new());
+    add_members(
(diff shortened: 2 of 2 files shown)


sagiegurari:
@pskrgag thanks a lot. i'm not sure why i added sorting... anyhow i'm running the CI and if all is good, i'll merge and publish a new version.

pskrgag:
@sagiegurari  Thank you for reviewing!

Digging more into rust std docs, I found that `.dedup()` actually removes only  **consecutive** repeated  elements [1]. Maybe `.unique()` is more suitable for initial idea in fe5ca3aba1050d0bce39663e0bd270d2a7d83049?

I can send follow up PR, if needed. I am sorry, I am still newbie in rust :)

[1] https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup

sagiegurari:
@pskrgag ok, so maybe thats the reason :D thanks for checking. let me put unique there instead

PR #941, opened 2023-09-09
move crate_target_dirs from RustInfo to CrateInfo

xxchan:
Because it uses `cargo metadata`, which can be time consuming on first invoke. Otherwise RustInfo is fast to load, and CrateInfo already uses `cargo metadata`.

And it also sounds more reasonable.

### DIFF
--- a/src/lib/environment/mod.rs
+++ b/src/lib/environment/mod.rs
@@ -345,7 +345,7 @@ fn setup_env_for_duckscript() {
     envmnt::set("CARGO_MAKE_DUCKSCRIPT_SDK_VERSION", version);
 }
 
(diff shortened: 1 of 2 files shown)


sagiegurari:
@xxchan thanks for this. 
question: if i'm not wrong, both rust and crate info are invoked on startup, so what is the gain (sorry, didn't check too much)?

xxchan:
Because
```
skip_rust_env_info = false
skip_crate_env_info = true
```
https://github.com/risingwavelabs/risingwave/blob/8ef74ad37194848ab7c75c7d39d503ad1d64025f/Makefile.toml#L81-L82

😄

sagiegurari:
thanks @xxchan merged

PR #993, opened 2023-11-22
Add Fig Autocompletion to README.md

beeinger:


beeinger:
@sagiegurari
Ready!

sagiegurari:
@beeinger thanks for this and sorry for really late reply.
can you push the content to the content.md instead of the readme.md (which is generated and therefore your changes will be deleted)?

sagiegurari:
nvm, i've merged it and i'll move the content to the right place.
thanks for the PR :)

PR #996, opened 2023-12-04
Enable setting default profile by env variable

Johnabell:
It would be useful to be able to set a default profile by env variable. This will help to have a different profile as default when, for example, working in a docker container vs working on the host machine.

This PR add the support for this allowing users to set `CARGO_MAKE_DEFAULT_PROFILE` so all future invocations of `cargo make` will use the given profile. If this variable is not set, it will fall back to the current behaviour of using `development` as the default profile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2662,6 +2662,10 @@ command = "echo"
 args = [ "running in production profile" ]
 ```
 
+#### Default profile
+
+The default profile is `"development"`. However, this can be overridden by setting the `CARGO_MAKE_DEFAULT_PROFILE` environment variable.
+
 <a name="usage-profiles-built-in"></a>
 #### Built In Profiles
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -2662,6 +2662,10 @@ command = "echo"
 args = [ "running in production profile" ]
 ```
 
+#### Default profile
```
sagiegurari, line 2665: can you put this in the content.md instead of readme as readme is generated?
also please add an anchor like link above and a link to this anchor in nav.md

sagiegurari:
@Johnabell this is great thanks. just 1 minor comment regarding the documentation going into the wrong file.

sagiegurari:
@Johnabell decided to fix the docs for you... thanks for the PR

PR #1009, opened 2024-01-03
fix: fix vscode link in readme.md

ben1009:


ben1009:
mark to draft, notice the readm.md is genetated by `[tasks.generate-readme]`, will modify the template in later change

sagiegurari:
@ben1009 thanks. ya, the content.md is the source

PR #1020, opened 2024-01-13
Add condition support for decode info.

### LINKED ISSUE #1018: Decode map environment variable overwrites existing one even with env_not_set condition
Hi, thanks for making and maintaining cargo-make, I really like it! I think I've come across a potential bug with conditionals and the decode map feature for environment variables. Here's an example Makefile.toml:

```toml
[tasks.conditional-source]
command = "echo"
args = ["${SOURCE_TEST}"]
env = {
  SOURCE_TEST = {
    condition = { env_not_set = ["SOURCE_TEST"] },
    source = "${SOURCE_IN}",
    default_value = "${SOURCE_IN}",
    mapping = { "one" = "1" }
  }
}
```

When I run this and don't have `SOURCE_TEST` set it works as expected and the `one` gets mapped to a `1`. But when a `SOURCE_TEST` *is* defined it overwrites that one and again maps it.

```txt
$ SOURCE_IN="one" makers conditional-source
[cargo-make] INFO - Execute Command: "echo" "1"
1

$ SOURCE_IN="one" SOURCE_TEST="example" makers conditional-source
[cargo-make] INFO - Execute Command: "echo" "1"
1

# Instead I would expect to get:
[cargo-make] INFO - Execute Command: "echo" "example"
example
```

This is using version 0.37.5.

Bauke:
Fixes #1018.

### DIFF
--- a/src/lib/environment/mod.rs
+++ b/src/lib/environment/mod.rs
@@ -126,17 +126,24 @@ fn set_env_for_script(key: &str, env_value: &EnvValueScript) {
 }
 
 fn set_env_for_decode_info(key: &str, decode_info: &EnvValueDecode) {
-    let source_value = expand_value(&decode_info.source);
-
-    let mapped_value = match decode_info.mapping.get(&source_value) {
-        Some(value) => value.to_string(),
-        None => match decode_info.default_value {
-            Some(ref value) => value.clone().to_string(),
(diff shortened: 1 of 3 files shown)


sagiegurari:
@Bauke merging. thanks :)

PR #1041, opened 2024-02-19
Allowed multiple ignored watch patterns

Buzzec:
Allows for multiple ignore patterns while watching as is allowed [here](https://github.com/watchexec/cargo-watch/blob/f931767c766b86005509b42f2f582c8ae2fb6221/src/args.rs#L141).


sagiegurari:
@Buzzec thanks for the PR. looks good. just will validate the CI is not broken and i'll merge it.

sagiegurari:
@Buzzec i'll fix the tests. thanks a lot for the PR

PR #1060, opened 2024-03-27
fix: reset CARGO env var to fix wrong CARGO in commands

### LINKED ISSUE #1059: `CARGO` env variable does not get set when running commands
### Describe The Bug
The README describes 

> When defined with scripts (as opposed to commands), the CARGO environment variable will be defined for the requested toolchain.

Although this says it doesn't define it for commands, I think this should so that the behavior matches with running via `rustup` or `cargo +toolchain`.

### To Reproduce
`build.rs`:
```
use std::path::PathBuf;
use std::env;

fn main() {
    panic!("{:#?}", env::var("CARGO").map(PathBuf::from).unwrap());
}
```

`Makefile.toml`:
```
[tasks.nightly-build]
toolchain = "nightly"
extend = "build"
```

These commands panic with stable toolchain path:
`cargo build`
`rustup run stable cargo build`
`cargo make build`
`cargo make nightly-build`

These commands panic with nightly toolchain path:
`cargo +nightly build`
`rustup run nightly cargo build`

The issue here is that `cargo make nightly-build` should print the nightly toolchain path. Something in cargo-make is preventing it from doing that.

wmmc88:
fixes #1059

### DIFF
--- a/src/lib/toolchain.rs
+++ b/src/lib/toolchain.rs
@@ -148,5 +148,5 @@ pub(crate) fn get_cargo_binary_path(toolchain: &ToolchainSpecifier) -> Option<St
 }
 
 pub(crate) fn remove_rust_env_vars() {
-    envmnt::remove_all(&vec!["RUSTC", "RUSTDOC", "RUSTFLAGS"]);
+    envmnt::remove_all(&vec!["CARGO", "RUSTC", "RUSTDOC", "RUSTFLAGS"]);
 }
--- a/src/lib/toolchain_test.rs
+++ b/src/lib/toolchain_test.rs
@@ -124,16 +124,19 @@ fn get_cargo_binary_path_valid() {
 #[test]
 #[ignore]
 fn remove_rust_env_vars_existed() {
+    envmnt::set("CARGO", "bad0");
     envmnt::set("RUSTC", "bad1");
     envmnt::set("RUSTDOC", "bad2");
     envmnt::set("RUSTFLAGS", "bad3");
 
+    assert!(envmnt::exists("CARGO"));
     assert!(envmnt::exists("RUSTC"));
     assert!(envmnt::exists("RUSTDOC"));
     assert!(envmnt::exists("RUSTFLAGS"));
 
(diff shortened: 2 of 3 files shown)


src/lib/toolchain_test.rs
```diff
@@ -124,16 +124,19 @@ fn get_cargo_binary_path_valid() {
 #[test]
 #[ignore]
```
wmmc88, line 125: This test is ignored in CI. Any reason for this @sagiegurari

wmmc88:
`codecov/project` seems to be failing because it considers the `toolchain_test.rs` file as requiring tests

src/lib/toolchain.rs
```diff
@@ -148,5 +148,5 @@ pub(crate) fn get_cargo_binary_path(toolchain: &ToolchainSpecifier) -> Option<St
 }
 
 pub(crate) fn remove_rust_env_vars() {
-    envmnt::remove_all(&vec!["RUSTC", "RUSTDOC", "RUSTFLAGS"]);
+    envmnt::remove_all(&vec!["CARGO", "RUSTC", "RUSTDOC", "RUSTFLAGS"]);
```
sagiegurari, line 151: ok i remember this. this was to fix the cargo bug. good that you found it.

src/lib/toolchain_test.rs
wmmc88: hmm. I'm not entierly sure how `ci-flow` executes this particular test, but my guess is that since `env::set_var` isn't safe for multi-threaded environments and some other test executing in parallel will change the env var and that spuriously causes this test to fail.

sagiegurari:
sorry fir the delay. I'll meege and publish a new version mid next week

wmmc88:
> sorry fir the delay. I'll meege and publish a new version mid next week

friendly bump 🙂

sagiegurari:
@wmmc88 thanks a lot. looks good. i'm merging to dev branch, doing few checks, check other issues and than i'll release.

PR #1109, opened 2024-07-07
Fix --skip-init-end-tasks argument 

### LINKED ISSUE #1108: `--skip-init-end-tasks` doesn't seem to work
### Describe The Bug
With `tasks.init` defined, even if we call `cargo make --skip-init-end-tasks empty`, the init task will be called.

I am not sure if I am understanding the `--skip-init-end-tasks` argument incorrectly, or there is a bug. If it's a misunderstanding, I believe we should change the documentation for the `--skip-init-end-tasks` argument: "If set, init and end tasks are skipped". If it's a bug, we need a fix.

### To Reproduce

On Linux with bash:
```bash
cargo init hello-world && cd hello-world && echo "tasks.init.script = \"echo init\"" > Makefile.toml && cargo make --skip-init-end-tasks empty
```

The actual output is:
```
$ cargo init hello-world && cd hello-world && echo "tasks.init.script = \"echo init\"" > Makefile.toml && cargo make --skip-init-end-tasks empty
    Creating binary (application) package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
[cargo-make] INFO - cargo make 0.37.13
[cargo-make] INFO - Calling cargo metadata to extract project info
[cargo-make] INFO - Cargo metadata done
[cargo-make] INFO - Project: hello-world
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: empty
[cargo-make] INFO - Running Task: init
init
[cargo-make] INFO - Build Done in 0.28 seconds.
```

I guess we are not supposed to see the "Running Task: init" log and the "init" echo?

### Error Stack

N/A

### Code Sample

```toml
tasks.init.script = "echo init"
```

This problem is important to me, as I plan to use the init task to add "-D warning" to the `RUSTFLAGS` environement variable for the CI environment. However, ff `--skip-init-end-tasks` doesn't work, the init task will be run another time when a new cargo-make process is forked due to `tasks.<task_id>.run_task.fork` being `true`. As a result, 2 duplicate "-D warnings" will be added to `RUSTFLAGS`, and cause recompilation.

EDIT: move the `--skip-init-end-tasks` argument before the empty target in To Reproduce section, and update the output.

06393993:
Fix #1108.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## CHANGELOG
 
+### v0.37.14
+
+* Fix: fix the --skip-init-end-tasks argument #1108
+
 ### v0.37.13 (2024-07-07)
 
 * Enhancement: New condition_script_runner_args attribute #1081
--- a/src/lib/cli.rs
+++ b/src/lib/cli.rs
@@ -149,14 +149,18 @@ fn run(cli_args: CliArgs, global_config: &GlobalConfig) {
             &env_info.crate_info,
         );
     } else if cli_args.print_only {
-        cli_commands::print_steps::print(
+        if let Err(e) = cli_commands::print_steps::print(
+            &mut std::io::stdout(),
             &config,
             &task,
             &cli_args.output_format,
             cli_args.disable_workspace,
             cli_args.skip_tasks_pattern,
             &env_info.crate_info,
-        );
+            cli_args.skip_init_end_tasks,
+        ) {
+            error!("Failed to print steps: {}", e);
+        }
     } else {
         runner::run(
             config,
--- a/src/lib/cli_commands/diff_steps.rs
+++ b/src/lib/cli_commands/diff_steps.rs
@@ -8,7 +8,7 @@
 mod diff_steps_test;
 
 use crate::command;
-use crate::execution_plan::create as create_execution_plan;
+use crate::execution_plan::ExecutionPlanBuilder;
 use crate::io::{create_file, delete_file};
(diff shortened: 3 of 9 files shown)


06393993:
It seems to me that all integrations tests are ignored, so I didn't add any integration tests for this feature. Please let me know if I am missing anything. Thanks.

sagiegurari:
tests are split to multi and single threaded. they are not ignored but you cant teat cargo make with cargo test. only via cargo make itself

06393993:
Added 2 integration tests. One thread safe, one not thread safe. Verified locally with:
* `cargo nextest run -E 'test(run_flow_skip_init_end_tasks)' --run-ignored all` and
* `cargo nextest run -E 'test(print_skip_init_end_tasks)' --run-ignored all`.

sagiegurari:
@06393993 looks good.

sagiegurari:
@06393993 can you fix the tests please? also moved it to a new dev branch

06393993:
I figured out the failure: for not thread-safe tests, we are setting the global logger in the new test, which may causes other tests to panic on `error!`. The fix is to reset the global logger after the newly added test completes.

06393993:
I have run ` cargo run --bin cargo-make --target-dir target/ci -- make ci-flow` locally on my machine to make sure it passes the currently failed github action step.

sagiegurari:
@06393993 merging. thanks a lot. i hope i'll release it this soon

PR #1121, opened 2024-07-22
[*.rs] Sort imports ; [*test*.rs] Use `assert_ne!` throughout

SamuelMarks:


src/lib/cli_commands/diff_steps.rs
```diff
@@ -7,15 +7,17 @@
 #[path = "diff_steps_test.rs"]
 mod diff_steps_test;
 
+use std::fs::File;
+use std::io;
+use std::io::{BufWriter, Write};
```
sagiegurari, line 12: why is this considered sorted?
i just put everything in a pile and let the formatter do it

SamuelMarks: 0. `std` libraries
1. third-party libraries
2. `crate` (first-party) libraries

Much more readable. I think I got the idea from `isort`. You can configure `rustfmt` to do this, see https://rust-lang.github.io/rustfmt/?version=v1.7.1&search=#StdExternalCrate%5C%3A

sagiegurari: ya i get that but i don't want to configure anything and stay consistent with other projects out there.
when ecosystem defaults change, i'll get those and fix as needed.
the assert_ne! parts are fine so suggest to have a PR just for that.

Issue #1110, opened 2024-07-08
`cargo-make` as library: task exection in asynchronous persistent queues


SamuelMarks:
### Feature Description
`cargo-make` is great. I've been using it since at least 2018. I have ideas that are too far outside the scope of `cargo-make` to be contributed back to `master`.

### Describe The Solution You'd Like
Can we expose `cli::descriptor::load` https://github.com/sagiegurari/cargo-make/blob/153fa22/src/lib/descriptor/mod.rs#L581-L586 ? - A bunch of other modules also? - Your `src/lib/cli.rs` uses a lot of private and `pub(crate)` `mod`ules.

### Code Sample
https://github.com/offscale/cargo-make/tree/publicise

---

I have ideas about making tasks asynchronous whilst retaining parallelizability by using a real database or Redis (valkey) to continue/resume/replay. But need to either fork your repo (sad) or get you to expose internals in order to experiment here.

Thanks for your consideration

PR #1127, opened 2024-07-26
Add `before_each_task` and `after_each_task` support

SamuelMarks:
WiP

---

Use cases this facilitates:
- Logging
- Subscriptions
- Distributed task execution (more details in #1110)
- More isolated tasks (think about when you use `beforeEach` and `afterEach` in your test-frameworks)

### DIFF
--- a/src/lib/execution_plan.rs
+++ b/src/lib/execution_plan.rs
@@ -19,6 +19,7 @@ use fsio::path::{get_basename, get_parent_directory};
 use glob::Pattern;
 use indexmap::IndexMap;
 use regex::Regex;
+use std::borrow::Cow;
 use std::collections::HashSet;
 use std::env;
 use std::path::Path;
@@ -481,8 +482,8 @@ impl<'a> ExecutionPlanBuilder<'a> {
             skip_tasks_pattern,
             skip_init_end_tasks,
         } = *self;
-        let mut task_names = HashSet::new();
-        let mut steps = Vec::new();
+        let mut task_names = HashSet::<String>::new();
+        let mut steps = Vec::<Step>::new();
         let default_crate_info = CrateInfo::new();
         let crate_info = crate_info.unwrap_or(&default_crate_info);
         let skip_init_end_tasks = skip_init_end_tasks || sub_flow;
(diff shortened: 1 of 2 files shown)


src/lib/descriptor/mod.rs
```diff
@@ -638,5 +638,13 @@ pub fn load(
         }
     }
 
+    for (name, _) in config.tasks.iter() {
+        match name.as_str() {
+            "before_each" => config.config.before_each_task = Some(name.to_string()),
+            "after_each" => config.config.after_each_task = Some(name.to_string()),
+            _ => {}
+        }
```
sagiegurari, line 646: what does this loop actually do? what does it mutate?
meaning are you setting config.before_each_task="before_each"? if so why loop? why not just configure it? why the naming magic?

src/lib/execution_plan.rs
```diff
@@ -549,6 +550,67 @@ impl<'a> ExecutionPlanBuilder<'a> {
...
+            Some(before_each) => handle_some(&before_each)?,
+        };
+        let scale: u8 = has_before_each + has_after_each;
+        if scale > 0 {
+            let before_and_after_each_len = scale as usize + 1;
+            let mut interspersed_steps =
+                Vec::<Step>::with_capacity(steps.len() * before_and_after_each_len);
+            let before_special = HashSet::from(["init", "init_task"]);
+            let end_special = HashSet::from(["end", "end_task"]);
+            interspersed_steps.extend(steps.into_iter().flat_map(|e| -> Vec<Step> {
```
sagiegurari, line 591: whats 'e'? can we give a more readable name?

src/lib/execution_plan.rs
```diff
@@ -549,6 +550,67 @@ impl<'a> ExecutionPlanBuilder<'a> {
...
+                let mut _steps = Vec::<Step>::with_capacity(before_and_after_each_len + 1);
+                if before_special.contains(e.name.as_str()) {
+                    if has_after_each == 1 {
+                        _steps.push(
+                            unsafe { after_and_before_each.last().unwrap_unchecked() }.to_owned(),
```
sagiegurari, line 596: do we really need unsafe code? seems like we are doing basic stuff. i know it will be more 'long/verbose' but... can we drop the unsafe?

src/lib/execution_plan.rs
```diff
@@ -549,6 +550,67 @@ impl<'a> ExecutionPlanBuilder<'a> {
...
+                        );
+                    }
+                } else if end_special.contains(e.name.as_str()) {
+                    if has_before_each == 1 {
+                        _steps.push(
+                            unsafe { after_and_before_each.first().unwrap_unchecked() }.to_owned(),
```
sagiegurari, line 602: same

src/lib/execution_plan.rs
```diff
@@ -549,6 +550,67 @@ impl<'a> ExecutionPlanBuilder<'a> {
...
+            None => 0,
+            Some(before_each) => handle_some(&before_each)?,
+        };
+        let scale: u8 = has_before_each + has_after_each;
+        if scale > 0 {
+            let before_and_after_each_len = scale as usize + 1;
+            let mut interspersed_steps =
+                Vec::<Step>::with_capacity(steps.len() * before_and_after_each_len);
+            let before_special = HashSet::from(["init", "init_task"]);
+            let end_special = HashSet::from(["end", "end_task"]);
```
sagiegurari, line 590: not sure about this. these names are configurable. why do we hardcode them?

src/lib/execution_plan.rs
```diff
@@ -549,6 +550,67 @@ impl<'a> ExecutionPlanBuilder<'a> {
...
+            interspersed_steps.extend(steps.into_iter().flat_map(|e| -> Vec<Step> {
+                let mut _steps = Vec::<Step>::with_capacity(before_and_after_each_len + 1);
+                if before_special.contains(e.name.as_str()) {
```
sagiegurari, line 593: maybe i'm not following this right but you are only pushing an additional task for the init/end tasks but not for other tasks instead of the opposite?

sagiegurari (changes requested):

src/lib/execution_plan.rs
SamuelMarks: The logic is:
- add a `before_each` (if exists) before each task; followed by an `after_each` (if exists) after each task

With two addenda:
- if the task is the `init` task and the `before_each` task exists, don't prepend the `before_each`
- else if the task is the `end` task and the `after_each` task exists, don't append the `after_each`

src/lib/execution_plan.rs
SamuelMarks: Can easily drop the `unsafe` and use `unwrap()` rather than `unwrap_unchecked()`. I added it because it's in a logic block that already confirms so it's proven safe. But sure, can replace this.

src/lib/descriptor/mod.rs
SamuelMarks: I couldn't figure out how to else get the `before_each_task` and `after_each_task` to be populated. Tried a bunch of solutions, this is the only one that seemed to make it accessible to future locations (specifically `src/lib/execution_plan.rs`).

src/lib/execution_plan.rs
SamuelMarks: Good point. Let me fix that

yonas:
Is this ready to be committed?

I recommend calling it `before_task` and `after_task`.

Issue #1081, opened 2024-05-08
Feature Request: Add a way to pass `args` to `script_runner` of a `condition_script`


wmmc88:
### Feature Description
I want to do the same thing as #1025, but for a `condition_script` instead of a normal `script` block. 

shebang support for condition scripts was added to resolve #987, but it looks like there isn't an equivalent to `script_runner_args` to pass args to the script runner selected by the shebang. I would like to pass args to the `rust-script` runner for the same reasons as in #1025.

### Describe The Solution You'd Like
Since the syntax for `condition_script` is slightly different that normal `script`, I'm not sure what the solution should look like. The simplest thing may by just to have a `condition_script_runner_args` field, but that is a little odd because there is no accompanying `condition_script_runner` field (i.e. using different runners only works via shebangs).

wmmc88:
Currently, I am working around this with a plugin, but I think this would be better suited inside of cargo-make itself. Especially since afaik, there is no way to run multiple different plugins.

```toml
[plugins.impl.rust-condition-script-cargo_make_current_task_initial_makefile_directory-substitution]
script = '''
# Parse task json into variables
taskjson = json_parse ${task.as_json}

# Convert backslashes to forward slashes
rust-driver-toolchian-path = replace ${taskjson.env.CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY} "\\" "/"

# Replace the ${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY} variable in the condition_script with the rust-driver-toolchian-path
taskjson.condition_script = replace ${taskjson.condition_script} "\${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY}" "${rust-driver-toolchian-path}"

# Reencode variables into json
taskjson = json_encode taskjson

# Run the invoking task, with the modified condition_script
cm_plugin_run_custom_task ${taskjson}
'''
```

sagiegurari:
@wmmc88 this is now available in the dev branch 0.37.13 and i would love your feedback as i did really bare minimum of tests.

sagiegurari:
@wmmc88 this is now officially release

PR #1132, opened 2024-07-31
Expand `condition_script_runner_args`

wmmc88:
#1081 added `condition_script_runner_args`, but `condition_script_runner_args` didn't expand env vars like `script_runner_args` did. This PR adds env expansion to `condition_script_runner_args`, and also adds tests for it.

Various bugfixes to tests were also fixed in order for `cargo make ci-flow` to pass on:
1. windows computers
2. environments that set `CARGO_HOME`


src/lib/runner.rs
```diff
@@ -361,6 +361,28 @@ pub(crate) fn run_task_with_options(
             None => (),
         };
 
+        let step = {
```
wmmc88, line 364: this block needed to be moved up so that expansion happens before condition validation (since condition validation calls the script runner with the args)

sagiegurari: i get it, but its not really what you want. you just want to expand.
this entire block also updated the current env.
so in case condition fails, the env is still updated. this is unexpected.

sagiegurari:
@wmmc88 thanks for the PR! would love if you see my comment

src/lib/runner.rs
wmmc88: Ok. Let me fix that.


For the evaluation of the condition script, is the correct behaviour to not use any env information defined in the current task when evaluating the condition for the current task?

README.md
```diff
@@ -1652,16 +1664,16 @@ These scripts use that value to create a new environment variable **`COMPOSITE_2
 * Load global environment variables defined in the **env.\[current profile\]** block.
 * Load global environment setup scripts defined in the **env_scripts** attribute.
 * **Per Task**
-  * Load environment files defined in the **env_files** attribute (relative paths are treated differently than global env_files).
   * Setup **per task** internal environment variables (see [Global](#usage-env-global) section).
+  * Load environment files defined in the **env_files** attribute (relative paths are treated differently than global env_files).
   * Load environment variables defined in the **env** block (same behavior as global env block).
```
wmmc88, line 1669: @sagiegurari This looks like the existing documentation was wrong, so I've corrected it to match what the code actually does (i.e. internal are expanded before env_files

wmmc88:
> @wmmc88 thanks for the PR! would love if you see my comment

@sagiegurari I've updated the PR so that there is no change in env when condition evaluation fails (and added a test to prevent possible future regressions)

wmmc88:
@sagiegurari friendly ping to review this :) the underlying bug is a blocker for me

sagiegurari:
@wmmc88 sorry for the delay! i'll go over it today

README.md
```diff
@@ -10,132 +10,144 @@
 
 > [Rust](https://www.rust-lang.org/) task runner and build tool.
 
-* [Overview](#overview)
```
sagiegurari, line 13: all changes in this file will be deleted as its auto generated.
nav changes go to nav.md (but please do'nt add that extra identation, not sure we need that new top level cargo-make item as we are already at the top)
and content changes go to content.md
both files are under docs/_includes/

src/lib/runner.rs
```diff
@@ -465,6 +466,11 @@ pub(crate) fn run_task_with_options(
                 };
             }
         } else {
+            // Restore env to state before expansion for condition evaluation
+            for (key, value) in env_before_current_task_env_expansion {
+                envmnt::set(key, value);
+            }
```
sagiegurari, line 472: the issue here is that new env vars are still stuck in the env. you only restore modified env vars.

sagiegurari (changes requested):
@wmmc88 everything looks pretty great just 2 main thinigs
1. docs - the file is generated so your changes will be deleted. lets try not to do the nav changes you did but only the content. you can add nav, but not add a new root please.
2. env setup before the condition and restoring it causes new env vars defined in the task to stick even if condition fails. we need to remove them.

src/lib/runner.rs
sagiegurari: @wmmc88 rethinking this whole thing, can we just not do all this env changes and only expand with what we have?
it is unexpected bahvior that env changes before condition is met.
if the condition needs some env vars, let them be defined in a dependency task as is.
this is much more readable and expected

wmmc88: I could rework the pr to not expand the env for the condition script, but I want to highlight that this means the some env will change before condition script and some will change after. Particularly, all the internal meta ends like `CARGO_MAKE_CURRENT_TASK*` will expand before condition, and the rest will expand after.

sagiegurari: i think what i'm saying is that for your PR, runner.rs does not need to change at all and control should be in the makefile author to setup env for conditions before hand

wmmc88: Before condition evaluation in runner.rs, there still needs to be a small change to call a function to expand args for condition_script_runner_args. 

If I understand you correctly, the `CARGO_MAKE_CURRENT_TASK_*` env vars will expand to the updated values for the current task (which matches existing behavior), but other env vars will not be updated (i.e. I remove the env expansion).

sagiegurari: if its just to call expand ya thats ok.

env is updated for all vars. no env var gets special invocation

README.md
wmmc88: I did not add nav changes intentionally. I think I just ran `cargo make ci-flow` locally, and this is what happened. I've reverted these changes and re-added the content changes to `content.md`.

src/lib/runner.rs
wmmc88: made change to only explicitly expand the condition-script-runner-args

wmmc88:
@sagiegurari I think I've resolved all your requested changes

sagiegurari:
Thanks a lot @wmmc88 for the PR. merged.
I'll probably release a new version at end of week.
sorry it took so long.

wmmc88:
> Thanks a lot @wmmc88 for the PR. merged. I'll probably release a new version at end of week. sorry it took so long.

Hi @sagiegurari! friendly bump for a release

PR #1157, opened 2024-09-04
Add test for condition_script_runner_args expansion

wmmc88:


src/lib/runner_test.rs
```diff
@@ -961,6 +963,96 @@ fn run_task_failed_condition_script_doesnt_change_env() {
...
+        allow_private: false,
+        skip_init_end_tasks: false,
+        skip_tasks_pattern: None,
+        cli_arguments: None,
+    };
+
+    let step = Step {
+        name: "test".to_string(),
+        config: Task {
+            install_crate: Some(InstallCrate::CrateInfo(InstallCrateInfo {
```
wmmc88, line 996: `install_crate` here doesnt do what i intended here since its run after the condition script is validated. I should just remove it.

wmmc88: @sagiegurari is there a better way to ensure that rust-script 0.35.0 is available for this test?

wmmc88:
@sagiegurari It looks like the test is failing only on MSRV rust on ubuntu. I (very hackily) forced some additional debug output and it looks like [rust-script fails to be installed w/ msrv rust](https://github.com/wmmc88/cargo-make/actions/runs/10714406740/job/29708042457?pr=2#step:7:2840). I see 2 issues with this:
1. `--locked` should be passed to [install for the scriptrunner](https://github.com/sagiegurari/cargo-make/blob/2c1bc771510ae9e115d23dd139204daba5d1d169/src/lib/scriptengine/rsscript.rs#L51)
2. `validate` should be set to true for pinstall for the rust-script runner](https://github.com/sagiegurari/cargo-make/blob/2c1bc771510ae9e115d23dd139204daba5d1d169/src/lib/scriptengine/rsscript.rs#L51). Right now it is false, so a failure to install rustscript is not correctly propogated upwards.

Do you agree that these changes make sense to make?

sagiegurari:
@wmmc88 yes both changes (lock and validate) makes sense. do you want to PR that?

wmmc88:
> @wmmc88 yes both changes (lock and validate) makes sense. do you want to PR that?

Yes. I'll update this pr with the changes

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: [stable, beta, nightly, 1.73.0]
+        rust: [stable, beta, nightly, 1.74.0]
```
wmmc88, line 18: rust-script's msrv is not 1.74:
```
error: failed to compile `rust-script v0.35.0`, intermediate artifacts can be found at `/tmp/cargo-install75pEMP`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  package `clap_lex v0.7.2` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.73.0
```

wmmc88:
> Yes. I'll update this pr with the changes

@sagiegurari The changes are implemented. MSRV was also bumped one version to 1.74 to accomodate for rust-script bumping its MSRV

src/lib/runner_test.rs
sagiegurari: you can create multiple tasks with dependencies.
so one task will do install and another task depending on the first, will do the condition.

sagiegurari:
@wmmc88 coming back to this. the tests are failing. wonder if you can take a look?

PR #1182, opened 2024-10-24
Add Zsh Completion Script Generation and Unit Tests

954alberto:
Summary: This PR introduces enhancements to the cargo-make tool by adding task completion script generation functionality per shell. Key updates include, in this case only Zsh was added:

Directory and File Management:

Automatically creates the ~/.zfunc directory if it does not exist.
Prompts users before overwriting the completion file if it already exists.
User Guidance:

Outputs clear instructions for enabling Zsh completions by modifying the ~/.zshrc file.
Testing Improvements:

Added unit tests for the completion generation function, including tests for user input handling and file creation.
Tests have been designed to clean up the test environment to avoid interference between test cases.

Ensures that the functionality can be tested without manually interacting with the terminal.


src/lib/cli_parser.rs
```diff
@@ -482,6 +500,10 @@ pub fn parse_args(
         let version_text = cliparser::version(&spec);
         println!("{}", version_text);
         Err(CargoMakeError::ExitCode(std::process::ExitCode::SUCCESS))
+    } else if let Some(shell) = cli_parsed.get_first_value("completion") {
+        // Call the function to generate completions
+        generate_completions(&shell);
+        Err(CargoMakeError::ExitCode(std::process::ExitCode::SUCCESS))
```
sagiegurari, line 506: why not return a result and than decide if success/fail based on that?

src/lib/completion.rs
```diff
@@ -0,0 +1,103 @@
...
+///    to facilitate adding support for other shell types.
+/// 3. **Enhanced Error Handling**: Provide more informative error messages for file operations.
+/// 4. **User Input Handling**: Ensure user input is trimmed and handled correctly.
+/// 5. **Testing**: Implement unit tests to verify the correct behavior of completion generation functions.
+
+#[cfg(test)]
+#[path = "completion_test.rs"]
+mod completion_test;
+
+pub fn generate_completions(shell: &str) {
```
sagiegurari, line 26: maybe return result?

src/lib/completion.rs
```diff
@@ -0,0 +1,103 @@
...
+    match shell {
+        "zsh" => {
+            if let Err(e) = generate_completion_zsh(None) {
+                eprintln!("Error generating Zsh completions: {}", e);
```
sagiegurari, line 30: maybe use logger? we have one that also stops on err

src/lib/completion.rs
```diff
@@ -0,0 +1,103 @@
...
+    let home_dir = std::env::var("HOME")?;
+    let zfunc_dir = format!("{}/.zfunc", home_dir);
+    let completion_file = format!("{}/_cargo-make", zfunc_dir);
+
+    if !Path::new(&zfunc_dir).exists() {
+        if let Err(e) = fs::create_dir_all(&zfunc_dir) {
+            eprintln!("Failed to create directory {}: {}", zfunc_dir, e);
+            return Err(Box::new(e));
+        }
+        println!("Created directory: {}", zfunc_dir);
```
sagiegurari, line 50: any reason we have println or errprintln and not logger?

src/lib/completion_test.rs
```diff
@@ -0,0 +1,90 @@
+use std::fs;
+use std::path::Path;
+use std::io::Cursor;
+
+#[cfg(test)]
+mod tests {
```
sagiegurari, line 6: i don't think this block is needed as this file is included only for cfg test to being with

src/lib/completion_test.rs
```diff
@@ -0,0 +1,90 @@
...
+    // Function to clean up test environment by removing the completion file
+    fn cleanup() {
+        let home_dir = std::env::var("HOME").expect("Failed to get HOME");
+        let completion_file = format!("{}/.zfunc/_cargo-make", home_dir);
+        if Path::new(&completion_file).exists() {
+            fs::remove_file(&completion_file).expect("Failed to clean up test file");
+        }
+    }
+
+    #[test]
```
sagiegurari, line 20: cargo make splits all tests to 2 parts
multi threaded
single threaded

your tests need to do a cleanup before each invocation so multi threaded won't work
to trigger them only in signle threaded (all tests run at the end...) you need to add ignore flag (see many tests of mine) which will postpone them to the single thread flow

sagiegurari (changes requested):
@954alberto this is great (sorry for late review).
would you mind also adding the relevant docs for this?
do it in the nav+content md files and not README.md which is generated from them.

src/lib/completion.rs
954alberto: no reason, you are right.

src/lib/completion.rs
954alberto: yes

src/lib/completion_test.rs
954alberto: blocked has been removed

src/lib/completion_test.rs
954alberto: Added #[ignore] to make the tests run single threaded

954alberto (commented):
Please have a look, it should be better now

sagiegurari:
@954alberto this is awesome. merging

PR #1187, opened 2024-11-13
New env_not condition

### LINKED ISSUE #1185: Enable Negative Conditions for Environment Variables
### Feature Description
Enable conditions for scripts where an environment variable is not a particular value.

There are cases where you might want to run a task only when an environment variable does not equal a particular value. This does not seem possible with the current options.

### Code Sample
```toml
[tasks.build-if-not-blue]
command = "cargo"
args = ["build"]
condition = { env_not = {  "COLOR" = "blue" } }
```

ayodejiige:
The PR includes a new `env_not` condition which will enable running tasks when environment variables are not equal to particular values. Resolves #1185


sagiegurari:
@ayodejiige thanks for the PR! i'll review and merge it. sorry for the HUGE delay!

sagiegurari:
@ayodejiige looks great. merging

PR #1190, opened 2024-11-19
Replace unsecure md5 with sha256 algorithm

marxin:
There are environments where broken cryptographically protocols are banned as part of company policy. MD5 should be considered cryptographically broken and unsuitable for further use. I suggest using more secure hashing algorithm.


sagiegurari:
thanks for the PR @marxin i'll review and merge

sagiegurari:
@marxin looks great. merging

marxin:
Thanks!

PR #1196, opened 2024-12-10
Platform name freebsd supports

stormshield-pj50:
This PR adds support for Freebsd platform name.

### DIFF
--- a/src/lib/types.rs
+++ b/src/lib/types.rs
@@ -22,6 +22,8 @@ pub fn get_platform_name() -> String {
         "windows".to_string()
     } else if cfg!(target_os = "macos") || cfg!(target_os = "ios") {
         "mac".to_string()
+    } else if cfg!(target_os = "freebsd") {
+        "freebsd".to_string()
     } else {
         "linux".to_string()
     }


sagiegurari:
thakns @stormshield-pj50 but any chance you can also update the docs?
its under docs/content the md file (the readme is generated from it)

docs/_includes/content.md
```diff
@@ -308,12 +308,15 @@ It is also possible to define platform specific aliases, for example:
 linux_alias = "linux_my_task"
 windows_alias = "windows_my_task"
 mac_alias = "mac_my_task"
+freebsd_alias = "freebsd_my_task"
```
sagiegurari, line 311: my hunch is that a bit more is needed for this to work.
1. add the attributes in the task type
2. proxy to it when platform is freebsd

if i'm not wrong.
and therefore i would also add a test for it.
without it, i feel the change you didn't won't actually impact anything. could be wrong but don't think so.

stormshield-pj50: I'm not sure how to add a relevant test for that but please take a look at my new commit and tell me if it's enough. Thank you.

PR #1211, opened 2025-01-02
Enable LTO and codegen-units = 1 optimization

zamazan4ik:
Resolves https://github.com/sagiegurari/cargo-make/issues/1202

Also, I decided to enable `codegen-units = 1` additionally in the Release profile since it allows us to squeeze even more optimizations with LTO.


sagiegurari:
@zamazan4ik thanks. will check it (after i fix the build)

sagiegurari:
@zamazan4ik looks good. merging.

PR #1218, opened 2025-01-18
fix zsh task completion for makers

kachick:
Sorry if I am wrong, when I reading https://github.com/sagiegurari/cargo-make/pull/1182, I have questions for the change.

I have just referred https://github.com/zsh-users/zsh-completions/blob/a7f01622f7bc6941d1c6297be6995fe1bbc9d4de/zsh-completions-howto.org#getting-started, not yet tested actual behaviors.

### DIFF
--- a/docs/api/src/cli/completion.rs.html
+++ b/docs/api/src/cli/completion.rs.html
@@ -190,12 +190,12 @@
     }
 
     <span class="kw">let </span>completion_script = <span class="string">r#"
-#compdef cargo make cargo-make
+#compdef 'cargo make' 'cargo-make make' makers
 
 _cargo_make() {
     local tasks
     local makefile="Makefile.toml"
-    
+
     if [[ ! -f $makefile ]]; then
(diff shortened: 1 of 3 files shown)


docs/api/src/cli/completion.rs.html
```diff
@@ -190,12 +190,12 @@
     }
 
     <span class="kw">let </span>completion_script = <span class="string">r#"
-#compdef cargo make cargo-make
+#compdef _cargo_make makers
```
kachick, line 193: I don't understand how to apply this for sub commands such as `cargo make` and `cargo-make make`.

kachick:
Sorry for making noise, closing because of I don't have enough knowledge in this context. And the old [bash completion](https://github.com/sagiegurari/cargo-make/blob/8a354723f9009a4454e5da806e63903f8fd9805b/extra/shell/makers-completion.bash) is still working better in my zsh.

PR #1301, opened 2025-12-08
workspace: fix arg parsing for args with whitespace

AstralZebulon:
CLI args with whitespace were not being perserved when creating the workspace task script, so that a single arg was being split into multiple args for the workspace member's task run.

 For example
   `--feature "f1 f2"`
was being seen as
   `--feature f1 f2`

The result was an error because the `f2` was not seen as associated with the `--feature` flag anymore.


AstralZebulon:
Should I file a bug for this PR?

